Issue #47, opened 2014-08-31
Windows find-definition doesn't have proper path output


zekesonxx:
`RUST_SRC_PATH=C:/Users/Owner/src/rust/src`
`main.rs:47   let mut cwd = os::getcwd();`

``` text
$ racer find-definition 47 17 src/main.rs
MATCH os,1,0,\Users\Owner\src\rust\src\libstd\os.rs,Module,\Users\Owner\src\rust\src\libstd\os.rs
```

PR #81, opened 2014-12-01
Fix RUST_SRC_PATH on Windows

GigaPatches:
Windows uses a semi-colon (`;`) rather than a colon (`:`) for separating paths due to drive letters. Currently, racer changes a path like `'C:\rust\src'` into `['C', '\rust\src']` (see #47). This may work or it may not (I don't know the rules for paths without drive letters, but it does work for some situations). 

I added a check for `#[cfg(windows)]` and `#[cfg(unix)]` to switch between using `;` and `:`.

### DIFF
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -22,6 +22,11 @@ use std::{str,vec};
 use std::iter::Iterator;
 use std;
 
+#[cfg(unix)]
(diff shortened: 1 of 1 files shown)


defuz:
Using different formats for different OS looks strange. Why not use `;` in all cases?

phildawes:
Because on unix the traditional path separator is ':'

PR #105, opened 2015-01-05
Fix compilation error for assoc types in the Iterator traits

kui:
This error is the following:

```
$ make
rustc -O -o bin/racer src/main.rs
src/racer/codeiter.rs:22:10: 22:32 error: wrong number of type arguments: expected 0, found 1
src/racer/codeiter.rs:22 impl<'a> Iterator<(uint, uint)> for StmtIndicesIter<'a> {
                                  ^~~~~~~~~~~~~~~~~~~~~~
```

This is caused by <https://github.com/rust-lang/rust/pull/20441>.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,6 @@
 #![feature(macro_rules)]
 #![feature(phase)]
+#![feature(associated_types)]
 #[phase(plugin, link)] extern crate log;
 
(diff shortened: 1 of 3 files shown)


kui:
cc: #105

phildawes:
Awesome - thanks very much

PR #106, opened 2015-01-05
Make racer friendly to vim package managers

bbatha:
Move racer.vim to expected location for pathogen, vundle, and neobundle.

Update README with instructions for using Vundle and NeoBundle


phildawes:
Cool - thanks v much!

PR #116, opened 2015-01-18
Change all slice methods to [..] 

tafia:
Hi,

Slice functions such as 'as_slice', 'slice_from', 'slice_to' and 'slice' are supposed to be deprecated in favor of range patterns [], [x..], [..x], [x..y] respectively.
This pull requests proposes all the necessary changes.
With the last rust, it reduces the number of warnings.

4 tests fails, as they do in master as well.

Just let me know if it helps or if I must make some more changes !

Johann

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,7 +18,7 @@ pub mod racer;
 #[cfg(not(test))]
 fn match_fn(m:Match) {
     let (linenum, charnum) = scopes::point_to_coords_from_file(&m.filepath, m.point).unwrap();
-    if m.matchstr.as_slice() == "" {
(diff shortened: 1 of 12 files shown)


phildawes:
Thanks @tafla, this is awesome!
Do you know if the recommended way to convert from String to &str is &*mystr or &mystr[] now? I've been doing the former, but I keep finding myself out of touch!

tafia:
No I'm not sure I feel &mystr[] better, in line with slices.

Thanks for the integration this is great !
I am not sure I've understood all the code, in particular sometimes you use byte sometimes chars ... 

Anyway, you did a great job and it'll help all of us !!

PR #117, opened 2015-01-20
Review codecleaner: iterators/chars instead of while/bytes

tafia:
Hi,

I did a review of codecleaner.

The main changes are :
- change byte to chars in order to improve readability (i am not sure if performance is better or not, but i think this might be marginal)
- change while loop to iterators. It feels more 'rust' this way. I think perf will be better (not that it mattered that much ... but why not !)
- change ifs to match: again feels more 'rust' and improves readability (to me at least)

I just tried to replicate the logic without having the very global picture. I trusted the tests to check if all is OK. 

Let me know if it helps or not and if i can continue on the other files.

### DIFF
--- a/src/racer/bench.rs
+++ b/src/racer/bench.rs
@@ -1,0 +1,23 @@
+extern crate test;
+
+use std::io::fs::PathExtensions;
+use std::os::getenv;
+use racer::codecleaner::code_chunks;
+use std::io::File;
+use self::test::Bencher;
+
+#[bench]
+fn bench_code_chunks(b: &mut Bencher) {
+    
+    let mut src_path = Path::new(&getenv("RUST_SRC_PATH").unwrap()[]);
+    src_path.push("libcollections");
+    src_path.push("bit.rs");
+    
+    let src = &File::open(&src_path).read_to_string().unwrap()[];
+    
+    b.iter(|| {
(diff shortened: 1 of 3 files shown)


phildawes:
Hi @tafia,

I agree this looks nicer, thanks. 
However the reason codecleaner uses bytes is to avoid utf8 decoding (which the Chars iterator does), so I'd want to make sure the performance is acceptable before merging this patch.
Also I'm not sure it is correct - does chars().enumerate() return an iterator that enumerates the byte positions or just the index of the character?

Cheers,

Phil

tafia:
I see. Let me write some benchs before. 
chars().enumerate() indeed return char index and not byte index.
The use of byte instead of char is just for performance reason or there are some scenarios where chars might be wrong ?

phildawes:
All the rust string functions take a byte position instead of a char position e.g. str.char_at(pos)

(they do this because otherwise e.g. char_at() would take linear time instead of constant time as it would need to scan all the characters before the position)

tafia:
Hi,
I have pushed some changes. In particular it reverts back to bytes.
Bench (i'll do another separate pull request for it) is now better than master and the code look simpler to me.
Again, let me know.

tafia:
Included in [request 123](https://github.com/phildawes/racer/pull/123)

PR #120, opened 2015-01-21
emacs: get racer-rust-src-path from getenv

geovex:
Gwt racer-rust-src-path from RUST_SRC_PATH environment variable if set


phildawes:
Cool, thanks @geovex

PR #121, opened 2015-01-22
Add bench module

tafia:
This adds a new bench.rs file, similar to test.rs.
It contains only one bench now but it is supposed to be extended.
Like test, this file is compiled only when testing and won't affect release build.


tafia:
Included in [request 123](https://github.com/phildawes/racer/pull/123)

Issue #126, opened 2015-01-29
Build fails on Jan 28th (and later) builds


Shchvova:
`rustc --version`
`rustc 1.0.0-nightly (c5961ad06 2015-01-28 21:49:38 +0000)`

`cargo build --release`
fails with error: 
```
src/racer/util.rs:81:13: 81:38 error: use of unmarked library feature
src/racer/util.rs:81 let s = std::rt::backtrace::write(&mut m)
```
replacing body `get_backtrace` with  `return "backtrace".to_string();` helps and autocomplete works, but it is hardly a solution.

ebfe:
The underlying problem seems to be that std::rt::backtrace is missing a stability attribute. Commenting out get_backtrace in util.rs and the single call to it in nameres.rs:756 shouldn't have any ill effects, as it its only used for debug output.

Shchvova:
Thanks. That is what I basically did :+1:

PR #127, opened 2015-01-29
Making Racer compile with newest Rust by removing unchecked debug calls

Shchvova:
As suggested in issue #126, backtrace is just breaking build now.


phildawes:
Thanks @Shchvova

tafia:
Hi
What is the purpose to keep commented code ?

I really think that commented lines won't have much attention anyway. That was the case with most projects I worked on, at the end it confuses readers because they think it is still a work in progress. 

If this particular piece of code if very important and the comment is just an emergency workaround, then we should probably have some `FIXME` comments (which on contrary may gain some attention), else i think git is here for us if we really want to see that particular lines.

What do you think ?

Shchvova:
I will uncomment/change them when runtime APIs would be stabilized.

Issue #41, opened 2014-08-13
Add a way to get the signature of a function


emberian:
For example, `racer signatureof fully::qualified::name` might print `fn(int) -> Box<String>`. Would also want this to work on methods, ideally.

c-cube:
This would be very nice to have. A similar project (in OCaml) binds this feature to `<Leader> t`, and it is extremely useful in presence of type inference. Almost as useful as completion imho.

llogiq:
What about `where` and generics?

Something like
```
BOUND 'a
BOUND T
RETURNS MyStruct<int,String> 
ARG foo: int 
ARG bar: Vec<String>
ARG baz: F
ARG frob: &'a T
WHERE F: FnOnce(&T) -> bool
```
perhaps?

phildawes:
Sounds reasonable to me

oblitum:
I'm seeing at the ReadMe that function signature is being shown. How is that happening? I don't have the same in Vim, any tip how to get it?

oblitum:
I see completion from command line is indeed returning more data than just identifier, so I guess the vim plugin was not updated to mirror that but the emacs part was?

PR #129, opened 2015-01-31
implement getting signature of function

krzat:
I'm trying to implement feature from issue #41.

This is not ready yet, I just wanted to ask something. I used find_definition_ to get declaration of the requested function, but in the complex case on the test, the contextstr is
"pub fn string_to_parser<'        'a ParseSess, source_str: String) -> Parser<'"
so few chars at the end are missing. 

Is this a bug of the find_definition_ function?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -120,12 +120,40 @@ fn find_definition() {
 }
 
 #[cfg(not(test))]
+fn signatureof() {
+        let args = std::os::args();
+    if args.len() < 5 {
+        println!("Provide more arguments!");
+        print_usage();
+        std::os::set_exit_status(1);
+        return;
+    }
+    let linenum = args[2].parse::<usize>().unwrap();
+    let charnum = args[3].parse::<usize>().unwrap();
+    let fname = &args[4][];
+    let fpath = Path::new(fname);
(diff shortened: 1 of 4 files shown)


krzat:
Nevermind, I could do this without find_definition_ (assuming supplied coordinates point at declaration already, which should happen if racer was used to get them).

phildawes:
Hi @krzat. I don't know sublimetext (I code rust in emacs), Can you use another key to trigger function signature getting? E.g. in intellij you use ctrl-p to get arguments

krzat:
I could probably do that, but pressing Ctrl+P every time user calls
function isn't very convenient. I will try observing '(' press and
determining if its function call,but its still not as good as inserting it
at the completion, IMO.

I was asking about JSON, because it would be easier to bake additional data
like this without breaking anything.

2015-02-02 7:15 GMT+01:00 Phil Dawes <notifications@github.com>:

krzat:
Anyway, the functionality from #41 seems to be achieved (except I added NAME to return name of the function too), so it's ready to merge (unless you have some problems with the code).

krzat:
closing for now, it won't work properly across files

PR #136, opened 2015-02-05
Update for path reform

renato-zannon:
This is necessary for racer to compile on a rustc newer than rust-lang/rust@8550bf74c1149eb54fc568bb53b78fd2b6caec98

The ideal strategy is to migrate racer to the new API, of course, but this at least keeps it compiling for now.


phildawes:
That's great, thanks Renato

PR #137, opened 2015-02-05
add snippets to method/function matches

krzat:
I had to rethink my approach because SublimeText (and perhaps other editors) have some serious limitations regarding autocompletion.

The purpose of this MR is to add snippet to the function match, so that editors can paste it on completion. That snippet has placeholders for function arguments, so it is easier to fill them. I really like this feature and would like to have it for Rust.

I used ??? string to separate snippet from the rest of the match string, because both snippets and context may contain commas, which would mess splitting. I'm open to suggestions if you have better idea how to handle this.

The snippet format seems to be supported by popular editors, i.e. SublimeText, Atom, Emacs and Vim. Other editors should be able to adjust it to their needs.

snippets
After selecting completion you should be able to switch between arguments with TAB.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,9 +17,25 @@ use racer::scopes;
 
 use racer::MatchType;
 
-use racer::method_info::MethodInfo;
-
 pub mod racer;
+
+#[cfg(not(test))]
+fn match_with_snippet_fn(m:Match) {
+    let (linenum, charnum) = scopes::point_to_coords_from_file(&m.filepath, m.point).unwrap();
+    if m.matchstr == "" {
+        panic!("MATCHSTR is empty - waddup?");
+    }
+
+    let snippet = racer::snippets::snippet_for_match(&m);
+    println!("MATCH {};{};{};{};{};{:?};{}", m.matchstr,
+                                    snippet,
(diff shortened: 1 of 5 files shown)


phildawes:
Hi @krzat, 
Nice! This seems like a good approach. I need to get a bit more familiar with snippets.
I'm afraid I can't merge this as-is because the ??? corrupts the other editor plugins output a bit. Could you maybe work this in as a separate command for the short term (e.g. complete-with-method-info or something)
Thanks!
Phil

krzat:
Done. I used this format: MATCH {};{};{};{};{};{:?};{}, where snippet is at second position.

Maybe I should just use JSON instead? I don't know how much the performance would suffer, but it would allow later for more customizability.

phildawes:
Hi @krzat, This is awesome! thanks for adding this. When I get some time I'll have a play with snippets on emacs.
Regarding json, I agree it would be a more flexible format. The advantage of the current csv-style format is that it's easy for each of the editors to parse without creating a dependency on a json library. I'm not sure if this is a good tradeoff long term; I'd want to know what was involved in parsing json in at least emacs and vim before switching though.

Thanks again,

Phil

PR #142, opened 2015-02-07
syntax::parse::token::InternedString usage updates

nyx:
I was trying to build against rustc 1.0.0-dev (a08504bb3 2015-02-07 09:55:14 +0000) and had to make some changes to src/racer/ast.rs.  Admittedly I am quite new to Rust but hope this is correct and helps!

PR #143, opened 2015-02-08
token::InternedString no longer has a .get() method

renato-zannon:
See rust-lang/rust#21505


renato-zannon:
Oops, just noticed #142 made the same changes! Closing this one, sorry for the noise

PR #144, opened 2015-02-08
Change std::os to std::env

tafia:
Silence several warnings in compilation


phildawes:
Thanks again @tafia !

PR #147, opened 2015-02-10
fix snippet errors for multiline method declarations

krzat:
There was with problem with functions spread over multiple lines. In that case contextstr contained only first line and caused panics when generating snippets.

This fixes this issue by searching for '{' and then removing unnecessary whitespace, so that contextstr fits inside single line.

### DIFF
--- a/src/racer/snippets.rs
+++ b/src/racer/snippets.rs
@@ -2,10 +2,14 @@ use syntax::ast::{FunctionRetTy, Method_};
 use syntax::ext::quote::rt::ToSource;
 use racer::ast::with_error_checking_parse;
 use racer::{Match, MatchType};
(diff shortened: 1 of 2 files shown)


phildawes:
Hi @krzat,

Rather than using contextstr could the snippet fetch the function declaration itself from the match?
(i.e. the match contains the file and the position of the function, so fetch the fn signature for parsing in the same way as typeinf::get_type_of_fnarg or typeinf::get_return_type_of_function)
'contextstr' in the match was a bit of an afterthought and might go away at some point.
Sorry if I'm wasting your time, 

Thanks!

krzat:
ok, done

phildawes:
Cool thanks!

PR #149, opened 2015-02-11
Move raw string management in string function

tafia:
It should be more efficient as the check for 'r' is done only once '"' is found
Also added some code cleaning


phildawes:
These look good, thanks! (merged)

PR #152, opened 2015-02-13
Fix build on current nightly

renato-zannon:
This is necessary due to the recent changes on the `std::env` module


phildawes:
Hi @renato-zannon 
Thanks v much for this. I merged #153 which also contains fixes for the same issue because it also contained other fixes.
Cheers,
Phil

PR #153, opened 2015-02-13
Fixes for rustc 1.0.0-nightly (3ef8ff1f8 2015-02-12 00:38:24 +0000)

Byron:
Even though this is a effective duplicate of #152, I use different syntax for the `env::args().collect()` calls.
Also there are some bugfixes, without which `racer` didn't work properly for me and `RustAutoComplete` (Sublime Text Plugin)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-#![feature(collections, core, io, os, path, rustc_private, std_misc, env, test)]
+#![feature(collections, core, io, path, rustc_private, std_misc, env, test)]
 
 #[macro_use] extern crate log;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -52,7 +52,7 @@ fn match_fn(m:Match) {
 
 #[cfg(not(test))]
 fn complete(match_found : &Fn(Match)) {
-    let args = std::env::args().map(|a| a.into_string().unwrap()).collect::<Vec<_>>();
+    let args: Vec<String> = std::env::args().collect();
     if args.len() < 3 {
         println!("Provide more arguments!");
         print_usage();
@@ -102,7 +102,7 @@ fn complete(match_found : &Fn(Match)) {
 
 #[cfg(not(test))]
 fn prefix() {
-    let args = std::env::args().map(|a| a.into_string().unwrap()).collect::<Vec<_>>();
+    let args: Vec<String> = std::env::args().collect();
     if args.len() < 5 {
         println!("Provide more arguments!");
         print_usage();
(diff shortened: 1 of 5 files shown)


Byron:
It seems besides the obvious fixes, there is another issue though, as simple completions now fail ungracefully:

```bash
RUST_BACKTRACE=1 target/racer complete 3 14 /Users/byron/Documents/dev/depot/src/keynote/res/rula/src/bin/reversor.rs
PREFIX 14,14,
thread '<main>' panicked at 'assertion failed: index < len', /Users/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-mac/build/src/libcollections/vec.rs:577
stack backtrace:
   1:        0x110a430a7 - sys::backtrace::write::h1864de2e29cfcfa3jWA
   2:        0x110a6abe3 - failure::on_fail::hb779450ca026c246YwJ
   3:        0x1109b5708 - rt::unwind::begin_unwind_inner::h252a44c757d4dd84eeJ
   4:        0x1109b5e71 - rt::unwind::begin_unwind_fmt::h6b3ebabbe58c4ea9KcJ
   5:        0x110a6a50e - rust_begin_unwind
   6:        0x110ab8757 - panicking::panic_fmt::haa2c8b833452fc7bSvw
   7:        0x110ab29f2 - panicking::panic::hdc8095395c2cbfb7Vtw
   8:        0x10fea2b42 - vec::Vec<T>::remove::h12763448109877615233
   9:        0x10fe810fd - racer::nameres::resolve_path::h5d65255f3a1f3cd6YNe
  10:        0x10fec376e - racer::complete_from_file::hc5d64468514f4bceZ2h
  11:        0x10fec7fce - complete::ha528e44216b55ed7lji
  12:        0x10fecb476 - main::h22bfb46b6f99fa0fBwi
  13:        0x110ae21a9 - rust_try_inner
  14:        0x110ae2196 - rust_try
  15:        0x110a6b62d - rt::lang_start::ha42b32fa3423fe20ZqJ
  16:        0x10fecb7bf - main
```

Byron:
The problem above is a plain old out-of-bounds issue, as the commandline parameter for the right column is not checked or clamped into range.
I think all that's needed is to either clamp it, or to fail with a an improve error message.

Byron:
Even though the latest commit fixes issues, there are plenty of tests missing to prevent regression. Also I don't know how things are supposed to be running, which is why it might be best if the maintainers would have a look (or let me know how it should be done).
Thank you

phildawes:
Hi @Byron ,
Thanks for this patch. I've applied because it fixes so much, and will take a look at the TODOs
Cheers,
Phil

PR #155, opened 2015-02-15
Improve matches

tafia:
Propose a better parsing in matchers (ie not necessarily separated with a simple space character)

It will do a better parsing of the first words in the blobs to ensure they match what is expected. They will also search for the `searchstr` directly and should prevent one parsing step.

In particular found that changing the first `complete_fn` test from

```Rust
fn apple

 to :
 
fn 
  apple
```
 broke the test.

NOTE: several matches have not been touched yet

### DIFF
--- a/src/racer/matchers.rs
+++ b/src/racer/matchers.rs
@@ -1,18 +1,12 @@
-use std::cell::{Cell};
-use std::{iter,option};
+use std::cell::Cell;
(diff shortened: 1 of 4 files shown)


phildawes:
This looks good, thanks for the patch.

tafia:
Thanks

PR #158, opened 2015-02-19
add Cargo lock and target/ folder to gitignore

bollu:
Edit the ```.gitignore``` to make sure that git doesn't see files generated by Cargo

Prevents git from seeing the ```target/``` folder where your executables end up.

Also, ignore the ```Cargo.lock``` file that cargo creates for synchronization.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ src/scopes
 !.travis.yml
 *tmpfile*
 *.racertmp
+Cargo.lock
+target/


tomjakubowski:
Cargo.lock shouldn't be ignored for packages that create only executables.

bollu:
I'm modifying racer to alternatively generate a library binding (a ```.so``` that lets you embed racer / provide a C FFI to be called). This _should_ be present since AFAIK even @Valloric and @phildawes are interested in this

phildawes:
Thanks @bollu 
N.B  racer did have a lib.rs for a bit; I've removed for now because the 'pass the code as a string' api didn't work. It'll be back!  (is in the git history if you want to take a look)

bollu:
@phildawes: I'm working on an FFI right now that I can use for Sublime text. The code is _very_ hacky right now.

I am to remove the whole environment variables thing and prevent racer from needing to reconstruct the entire "graph" from scratch. Ideally, you should be able to pass a ```RacerContext``` or something that can get incrementally updated. I'm working on that right now!

[Here](https://github.com/bollu/racer/commit/199a4db1c420d005d8acb48eec6804c9bd14a2d1) is the diff of the first commit I've made towards these goals.

As of now, the whole environment variables are hardcoded in. That's going to change - I'm planning on passing around an incrementally built ```struct``` to save the state, as well as the program input as a set of file paths(?) or strings(?). I'm not entirely sure how it should be structured.

Hope you're okay with me working on it! Thanks!

phildawes:
Hi @bollu!

FFI for sublime text sounds cool. Have you seen this?: 
https://mail.mozilla.org/pipermail/rust-dev/2014-August/011039.html

Re caching the 'graph', note that racer only parses the code statements it needs to complete a search, not the whole source tree.

I do think racer will need to start passing a context around, or maybe something in thread local storage. However I'm currently defering making any design changes until I know how best to interact with rustc for type inference. I'd like to have rustc do the heavy lifting for type inference rather than racer doing it's own, so I need to spend investigating how best to achieve this.

Cheers,

Phil

Issue #157, opened 2015-02-18
Racer spawns a large number of one-off threads... why?


renato-zannon:
On `src/racer/ast.rs`, there are multiple methods whose implementation looks something like this:

```rust
    let thread = Thread::scoped(move || {
        let stmt = string_to_stmt(s);
        let mut v = ImplVisitor { name_path: None, trait_path: None };
        visit::walk_stmt(&mut v, &*stmt);
        return v;
    });
    return thread.join().ok().unwrap();
```

Since this is using `Thread::scoped` (which makes the parent thread wait on the `thread.join()` call), there isn't any gain in performance here - on the contrary, there's a big performance loss. On my linux system, a call to `racer complete std::old_io::` spawns 108 threads, and spends almost half of its execution time on the `futex` syscall (thread synchronization).

On [a comment](https://github.com/phildawes/racer/blob/42c05827fba799641e0aa4d0a45f835bb23c1f82/src/racer/ast.rs#L800), there's a possible justification for this - handling `panic!`s from the parser, since Rust provides no way to "catch" `panic!`s. The problem with this justification is twofold - first, methods as the one above (which are by far the most common on the file I mentioned) aren't actually catching anything, since they `unwrap()` the final result of `thread.join()` - meaning that a `panic!` will be propagated to the parent thread anyway. Secondly, per rust-lang/rust#22435, "catching" panics at the border of a `scoped` thread won't be supported by the stdlib anymore - which means that racer won't be able to use this technique anymore in a few days.

I'm opening this issue so that we can discuss a possible solution for this - hopefully one that doesn't use threads for control flow :)

phildawes:
Hi Renato,
You are correct in thinking it was originally to catch panics. Racer has got much better at passing parsable blobs of code to libsyntax now so it's not needed and remains mainly because of my lazy copy-paste coding style! Feel free to fix if you want, or I'll repair it when rust-lang/rust#22435 hits.

BTW Thanks for pointing me at rust-lang/rust#22435 - I'll check out the discussion.

Cheers,

Phil

phildawes:
This is great, thanks Renato

PR #159, opened 2015-02-19
Fix build on current nightly (by removing threads from src/ast.rs)

renato-zannon:
This fixes the build on the current nightly, and also #157, by simply removing all calls to `Thread::scoped` from `src/ast.rs`. All tests still pass, and no regressions have been found after some initial testing.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-#![feature(collections, core, io, path, rustc_private, std_misc, env, test)]
+#![feature(collections, core, old_io, old_path, rustc_private, env, test)]
 
 #[macro_use] extern crate log;
 
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -16,7 +16,6 @@ use racer::Ty::{TyTuple, TyPathSearch, TyMatch, TyUnsupported};
 use syntax::ptr::P;
 use syntax::visit::Visitor;
 use racer::nameres;
-use std::thread::Thread;
 
 
 // This code ripped from libsyntax::util::parser_testing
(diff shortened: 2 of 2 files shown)


tafia:
For my information #157 raised some performance issue. Did you have tile to
make some benchmark to see if it changes something ?

renato-zannon:
Just did!

```
$ cat bench.sh
N=100
for i in $(seq 1 $N); do
  target/release/racer complete std::old_io:: >/dev/null
done

# Without the PR:
$ rustc --version
rustc 1.0.0-nightly (b63cee4a1 2015-02-14 17:01:11 +0000)

$ time ./bench.sh
real    0m18.654s
user    0m10.353s
sys     0m8.313s

$ cargo bench
# <snip>
test racer::bench::bench_code_chunks     ... bench:    200484 ns/iter (+/- 11952)
test racer::bench::bench_iter_stmts      ... bench:    478272 ns/iter (+/- 23615)
test racer::bench::bench_mask_comments   ... bench:    296777 ns/iter (+/- 11999)
test racer::bench::bench_mask_sub_scopes ... bench:    399283 ns/iter (+/- 15699)

# With the PR:

$ rustc --version
rustc 1.0.0-nightly (dfc5c0f1e 2015-02-18) (built 2015-02-19)

$ time ./bench.sh
real    0m11.323s
user    0m7.189s
sys     0m4.117s

$ cargo bench
# <snip>
test racer::bench::bench_code_chunks     ... bench:    198126 ns/iter (+/- 6881)
test racer::bench::bench_iter_stmts      ... bench:    483020 ns/iter (+/- 20938)
test racer::bench::bench_mask_comments   ... bench:    303417 ns/iter (+/- 11246)
test racer::bench::bench_mask_sub_scopes ... bench:    388178 ns/iter (+/- 14858)
```

So it seems that it didn't influence the benchmarks too much (maybe they don't hit this code path?), but there's a big difference on runtime.

That said, the main reason for this PR is that the behavior racer was counting on - using scoped threads to catch panics - isn't supported by rust anymore, and the code needed to be changed to reflect that anyway. The performance improvement is just a nice side effect :)

tafia:
Thanks !
I always like nice effects

tafia:
And for the benchmarks it is normal, there are benchmarking only some
specific functionalities (we could actually remove them and add yours, or a
rust only version,  which is more transversal)

phildawes:
Quick (relatively unscientific) test on a linux laptop:
```rust
let v = Vec::new();
v.<complete>
```

Before Renato's patch:
```
% time ./target/release/racer complete 2 2 /tmp/a.rs
real	0m0.635s
user	0m0.331s
sys	0m0.309s
```

After:
```
% time ./target/release/racer complete 2 2 /tmp/a.rs
real	0m0.323s
user	0m0.187s
sys	0m0.135s
```
Nice!

PR #164, opened 2015-02-25
Use the new Thread API to sandbox libsyntax use

bollu:
This allows RACER to continue processing even when libsyntax crashes.
This is vital if we wish to use RACER as say, a library in the future.

This fixes [the issue of racer crashing](https://github.com/phildawes/racer/issues/162)

I realize that this commit sort of undoes [this one where the threading code was removed](https://github.com/phildawes/racer/commit/f82eae11e5923b6cfc79f0ff89edd47bc87e91dd), but from what I read on the [issue tracker](https://github.com/phildawes/racer/issues/157), that code was removed since Rust's behaviour changed and allowed ```panic!(...)``` to leak between threads.

Since the leaking of ```panic!(...)``` does not happen, I think that this error handling _must_ be reintroduced, regardless of performance cost (yes, this does have some performance penalty)

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -17,6 +17,18 @@ use syntax::ptr::P;
 use syntax::visit::Visitor;
 use racer::nameres;
 
+/*
+since libsyntax panics when we send it a syntactically wrong file,
+we need to make sure that when it panics, it does not bring RACER
+down with it. This is doubly important when RACER is used as a library,
+since crashing a library through the FFI will also crash anything that uses
+the library. Therefore, to sandbox crashes, use Rust's threads that are able to do this.
+*/
+//for sending and receiving data between threads
+use std::sync::mpsc;
+//for thread::spawn (to spawn a new thread)
+use std::thread;
+
 
 // This code ripped from libsyntax::util::parser_testing
 pub fn string_to_parser<'a>(ps: &'a ParseSess, source_str: String) -> Parser<'a> {
@@ -795,27 +807,65 @@ impl<'v> visit::Visitor<'v> for EnumVisitor {
     }
 }
 
+
 pub fn parse_use(s: String) -> UseVisitor {
-    let cr = string_to_crate(s);
-    let mut v = UseVisitor{ident: None, 
-                                paths: Vec::new(),
-                                is_glob: false};
-    visit::walk_crate(&mut v, &cr);
-    return v;
+    let (tx, rx) = mpsc::channel();
+
+    thread::spawn(move || {
+        let cr = string_to_crate(s);
+        let mut v = UseVisitor{ident: None, 
+                                    paths: Vec::new(),
+                                    is_glob: false};
+        visit::walk_crate(&mut v, &cr);
+
+        //we need the unwrap since send() returns a Result.
+        //if there is an error, unwrap() will cause a panic
+        tx.send(v).unwrap();    
+    });
+
+    match rx.recv() {
+        Ok(s) => { return s; },
+        Err(_) => {
+            return UseVisitor{ident: None, 
+                              paths: Vec::new(),
+                              is_glob: false};
+        }
+    }
 }
 
 pub fn parse_let(s: String) -> Vec<(usize, usize)> {
-    let stmt = string_to_stmt(s);
-    let mut v = LetVisitor{ ident_points: Vec::new() };
-    visit::walk_stmt(&mut v, &*stmt);
-    return v.ident_points;
+    let (tx, rx) = mpsc::channel();
+
+     thread::spawn(move || { 
+        let stmt = string_to_stmt(s);
+        let mut v = LetVisitor{ ident_points: Vec::new() };
+        visit::walk_stmt(&mut v, &*stmt);
+
+        tx.send(v.ident_points).unwrap();
+    });
+
+    match rx.recv() {
+        Ok(ident_points) => ident_points,
+        Err(_) => Vec::new()
+    }
 }
 
 pub fn parse_struct_fields(s: String, scope: Scope) -> Vec<(String, usize, Option<racer::Ty>)> {
-    let stmt = string_to_stmt(s);
-    let mut v = StructVisitor{ scope: scope, fields: Vec::new() };
-    visit::walk_stmt(&mut v, &*stmt);
-    return v.fields;
+
+    let (tx, rx) = mpsc::channel();
+
+    thread::spawn(move || {
+        let stmt = string_to_stmt(s);
+        let mut v = StructVisitor{ scope: scope, fields: Vec::new() };
+        visit::walk_stmt(&mut v, &*stmt);
+
+        tx.send(v.fields).unwrap();
+    });
+
+    match rx.recv() {
+        Ok(fields) => fields,
+        Err(_) => Vec::new()
+    }
 }
 
 pub fn parse_impl(s: String) -> ImplVisitor {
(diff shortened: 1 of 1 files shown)


tafia:
Is thread the only way to handle a panic ?
Can't we have a unique thread wrapping it all ?

bollu:
@tafia: good point. But in that case, we'll have to manage the liveness of that thread (like, if that thread crashes, then we need to spin up a replacement thread). Plus, how would job control work? I'm not sure how you could use one thread to do this when (afaik) these procedures sometimes mutually recursively call each other.

tafia:
Yes
For library purpose, is the canonical way really to have a thread for each
and every function ?
I find it strange. Better have a wrapper for libraries while main directly
could have its own private thread.
No more nested calls.
It still feels a wrong

krzat:
Can I suggest encapsulating this into a function? Otherwise that's a lot of boilerplate to copy-paste. Also it would be easier to toggle it off later.

bollu:
hm, but the problem with that is the fact that the way the error handling works is different for different functions. It could be modularized however, I don't really think that's necessary. Although, I could do this. 

I'm kind of debating on the pros versus cons of this.

phildawes:
Hello!

Thanks for the patch @bollu, and the conversation. I've been flip-flopping on this one. It makes racer half the speed again:

before:
```
real	0m0.352s
user	0m0.203s
sys	0m0.147s
```

after:
```
real	0m0.628s
user	0m0.300s
sys	0m0.405s
```

However the issue of libsyntax crashing also affects some completions: racer is a quite liberal at the moment with what it passes to libsyntax. E.g. the following currently causes a crash:

```rust
fn main() {
    let v = Vec::  // <-- complete here
    let b = 3;
}
```

Whereas it used to complete fine and does again after this patch. (The reason is that while searching for "Vec", it matches the current line and attempts to parse it, but the statement overflows into the next line).

Ideally libsyntax would change so that it doesn't panic any more. I suspect this will happen at some point since this is the direction rust is going in, although it may not be soon.

In the meantime I'm thinking that we're going to need to apply this patch just to get racer working again in the presence of incomplete lines.

Cheers,

Phil

phildawes:
Now that I think about it I haven't actually investigated the possibility that there could be a way to use libsyntax without panics. I'll investigate tonight if I have time.

phildawes:
Oh duh, I think we're just aborting because of some code I cut-n-paste back in the early days!
```rust
pub fn with_error_checking_parse<F, T>(s: String, f: F) -> T where F: Fn(&mut Parser) -> T {
    let ps = new_parse_sess();
    let mut p = string_to_parser(&ps, s);
    let x = f(&mut p);
    p.abort_if_errors();   // <<<<<<<<<<<<<<----------------------- HERE!
    x
}
```
Will remove this tonight and see what happens

bollu:
@phildawes: don't we _need_ the ```abort_if_errors```? I was under the impression that the particular function must be called.

bollu:
I was testing [this file](https://github.com/bollu/subl-racer-profiler/blob/master/tests/errored_rust.rs) against RACER to see what happens when there are errors.

I removed the ```abort_if_errors()``` from the ```with_error_checking_parse(...)```.

Here's the output:
```
1 22 tests/errored_rust.rs
bogofile:1:7: 1:9 error: expected one of `;` or `as`, found `td`
bogofile:1 use s td::
                 ^~
thread '<main>' panicked at 'Box<Any>', /home/bollu/prog/rust/src/libsyntax/diagnostic.rs:92
PREFIX 11,11,
```

The stack trace is here:
```
thread '<main>' panicked at 'Box<Any>', /home/bollu/prog/rust/src/libsyntax/diagnostic.rs:92
stack backtrace:
   1:     0x7f6a369c4fe0 - sys::backtrace::write::hd237769f574cebd0KlC
   2:     0x7f6a369ed890 - panicking::on_panic::hd026f39bc548b60ftXL
   3:     0x7f6a3692e4f0 - rt::unwind::begin_unwind_inner::h5412fee8459176ffRBL
   4:     0x7f6a3734a920 - rt::unwind::begin_unwind::h11880206013635135768
   5:     0x7f6a372f09a0 - diagnostic::SpanHandler::span_fatal::h69a6eb53eef78d20sSE
   6:     0x7f6a3735c580 - parse::parser::Parser<'a>::expect_one_of::h8e9bf8acfc2488c8xRI
   7:     0x7f6a3735dc10 - parse::parser::Parser<'a>::expect::h93479551fce8f3293PI
   8:     0x7f6a373868f0 - parse::parser::Parser<'a>::parse_item_::hb75827bd2292b996zUN
   9:     0x7f6a37397e60 - parse::parser::Parser<'a>::parse_mod_items::h5395ab0c6c705996PjN
  10:     0x7f6a373a33b0 - parse::parser::Parser<'a>::parse_crate_mod::hf2728e1de9b80fd3FlO
  11:     0x7f6a37e9d280 - racer::ast::string_to_crate::closure.13957
  12:     0x7f6a37e9d0a0 - racer::ast::with_error_checking_parse::h9486131292660222995
  13:     0x7f6a37e9cff0 - racer::ast::string_to_crate::hbb393764668d2f3ahFa
  14:     0x7f6a37ed1af0 - racer::ast::parse_use::h5a175402a9a9e009YZb
  15:     0x7f6a37f4be80 - racer::matchers::match_use::h96b1d2939d885291Bzg
  16:     0x7f6a37f3b3e0 - racer::matchers::match_types::h9ff813c305aa9916bQf
  17:     0x7f6a37f39680 - racer::nameres::run_matchers_on_blob::heabf7272047f6282Abe
  18:     0x7f6a37f35a90 - racer::nameres::search_scope::h83058985f8204811E0d
  19:     0x7f6a37f3d010 - racer::nameres::search_local_scopes::h893dc46214980654vfe
  20:     0x7f6a37f336c0 - racer::nameres::resolve_name::hc4ad6c0d77ef56beTCe
  21:     0x7f6a37f207b0 - racer::nameres::resolve_path::h3320f02c86c3d3f5OJe
  22:     0x7f6a37f5cf20 - racer::complete_from_file::h421ec1888af1fb81pGh
  23:     0x7f6a37f618d0 - complete::hdf9353fba9529135LWh
  24:     0x7f6a37f65660 - main::h9f9740ef5524604819h
  25:     0x7f6a36a64dd0 - rust_try_inner
  26:     0x7f6a36a64dc0 - rust_try
  27:     0x7f6a369ef550 - rt::lang_start::hf1a44dd833e003b5tRL
  28:     0x7f6a37f65c20 - main
  29:     0x7f6a36098dd0 - __libc_start_main
  30:     0x7f6a37e6f670 - <unknown>
  31:                0x0 - <unknown>

```

looking at the source code, it crashes [here, with a call to ```fatal()```](http://doc.rust-lang.org/src/syntax/parse/parser.rs.html#428). This then goes on to finally call ```panic!``` within [SpanHandler](http://doc.rust-lang.org/src/syntax/diagnostic.rs.html#90)

So the ```abort_if_errors()``` codepath has _no impact_ on the crash here. So removing it has no impact.
we _do_ need threads to sandbox this. Or, someone needs to rewrite ```libsyntax```. But I'm not sure when that's happening :)

phildawes:
Hello!,

Thanks for the test case, I agree that a racer lib shouldn't panic. 

If we cannot get our libsyntax use to be panic-free then a better solution to wrapping every libsyntax call could be to wrap at a higher level (e.g. in the top-level function in the racer lib), for performance reasons. There's not much point avoiding the cost of spawning a process and then burning up to 10x that amount of time in thread creation and locking.

Actually that leads me to a question: What's your main motivation for doing a sublime text FFI - is it performance or something else?

Thanks,

Phil

krzat:
We could also keep one thread for executing libsyntax and create new one each time the previous panics.

bollu:
hm, that seems reasonable-ish. But again, what happens if these are called recursively?

If A calls B, and A was running on thread T, then what happens? to we spawn a new thread T1? how do we know when to spawn a new thread?

bollu:
it's a mixture of multiple things - FFI, more complex completion support, the need to not parse strings every time we pass through Racer, and just provide a better user interface.

I'm planning on packaging RACER as a library next as part of [my GSOC proposal](http://bollu.github.io/). So this is the first step.

phildawes:
Hi @bollu, I'm afraid the link to the GSOC proposal isn't working for me now. I think it was working earlier today.

phildawes:
FYI: http://internals.rust-lang.org/t/changing-libsyntax-to-use-result-instead-of-panic/1670

bollu:
@phildawes: I updated the link - should work now - [Here's the URL](http://bollu.github.io/gsoc/gsoc.html)

phildawes:
Hi @bollu, I had a look through the GSOC proposal. It would be cool to have you working on racer!

A couple of things:

- Calling racer 'completely unoptimized' isn't very accurate (and a little disrespectful), I and others have put many hours into performance. It does some pretty sophisticated searching and type inference across crates and returns results in a fraction of a second.

- It seems that this proposal places heavy emphasis on caching and serializing to a binary format for performance, but you are not specific about what you plan to cache and serialize, or what the performance goals are. 

Racer in its current state doesn't have much (if anything) to cache between invocations - it does not compile whole files or crates into ast and it doesn't build a type graph. Do you intend to change racer so that it does this (e.g. by interfacing somehow with rustc and compiling crates globally into typed ast)? This would be significant work so it is worth elaborating on this.

phildawes:
FYI: https://github.com/rust-lang/rust/pull/23857

phildawes:
Closing as libsyntax has now been patched to remove panics and racer uses this. Further panics should be considered a bug and fixed in racer/libsyntax. Cheers!

PR #166, opened 2015-02-27
fix(rustup): adjust to changed TyPath variant

Byron:
The fix was simple, as variant fields just had to be swapped.


phildawes:
Apologies, I made the same fix. Thanks v much for the patch

PR #169, opened 2015-03-06
Use #![cfg_attr()] for declaring the exit_status and test features

daboross:
This puts `#![feature(exit_status)]` under `#![cfg_attr(not(test))]`, and ``#![feature(test)]` under `#![cfg_attr(test)]` in order to be rid of all unused feature warnings.

`#![cfg_attr(..., ...)]` works just like `![cfg(...)]`, except for declaring attributes.


daboross:
This build does fail, but it fails in the same place `master` does - if you look at the log it does successfully fix the feature not used warnings.

phildawes:
Nice! Thanks v much @daboross

daboross:
No problem!

PR #177, opened 2015-03-20
updated calls for libsyntax changes

vincom2:
`Parser::parse_item` and `Parser::parse_stmt` lost a parameter each. Also, `Parser::parse_stmt` now returns an `Option`. Changes made were ripped from the updated `libsyntax::util::parser_testing`, just like the original.


phildawes:
Cool, thanks @vincom2 !

PR #180, opened 2015-03-22
Use new io APIs

### LINKED ISSUE #179: cannot compile on rustc 1.0.0-nightly (ecf8c64e1 2015-03-21) (built 2015-03-22)
There are a couple of problems. We have lots of undeclared uses of Path. 

```
src/racer/ast.rs:338:50: 338:54 error: use of undeclared type name `Path`
src/racer/ast.rs:338 fn resolve_ast_path(path: &ast::Path, filepath: &Path, pos: usize) -> Option<Match> {
```

If I use std::path::Path everywhere, I still get an error:

```
trait `core::marker::Sized` is not implemented for the type `[u8]` 
```

This is on Ubuntu 14.04.

oakes:
This transitions racer to use all the new io APIs to fix the latest build errors. Fixes #179.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-#![feature(collections, core, old_io, old_path, rustc_private)]
+#![feature(collections, core, rustc_private, str_char)]
 #![cfg_attr(not(test), feature(exit_status))] // we don't need exit_status feature when testing
 #![cfg_attr(test, feature(test))] // we only need test feature when testing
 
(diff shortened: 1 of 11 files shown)


Ell:
Tried this out and this seems to fix all the errors I was getting from trying to compile using nightly as of 3-22-2015. +1

phildawes:
Awesome, thanks @oakes !

PR #190, opened 2015-04-08
Show signature of functions in vim popup menu

dan-t:
e.g. for the 'Vec' type and the string 'push' the
popup menu previously only showed:

   push
   push_all

Now the popup menu will show:

   push     f pub fn push(&mut self, value: T)
   push_all f pub fn push_all(&mut self, other: &[T])

Currently functions get a 'f' marker and struct members a 'm'.

### DIFF
--- a/plugin/racer.vim
+++ b/plugin/racer.vim
@@ -100,8 +100,17 @@ function! racer#GetCompletions()
     let out = []
     for line in lines
        if line =~ "^MATCH"
-           let completion = split(line[6:], ",")[0]
-           let out = add(out, completion)
+           let line_parts = split(line[6:], ",")
+           let name = line_parts[0]
+	   let kind = ""
+           let type = ""
+	   if line_parts[4] == "StructField"
+	       let kind = "m"
+           elseif line_parts[4] == "Function"
+	       let kind = "f"
+               let type = join(line_parts[5:], ",")
+	   endif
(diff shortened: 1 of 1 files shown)


phildawes:
Hi @dan-t. I don't know much about the vim plugin - did you look at 'racer#GetExpCompletions()'? Does this supercede that?
Thanks!

dan-t:
Oh, I haven't been aware of 'racer#GetExpCompletions()', so yes, this seems to be the same thing.

phildawes:
BTW, I don't use vim - if you feel that GetExpCompletions should be the default feel free to change this.

Thanks!

dan-t:
Hi Phil,

On Wed, Apr 08, 2015 at 11:41:18PM -0700, Phil Dawes wrote:
> BTW, I don't use vim - if you feel that GetExpCompletions should be the default
> feel free to change this.

If you're looking at the screenshots, you will see the different
displaying of GetCompletions and GetExpCompletions.

Further GetCompletions doesn't start a python interpreter for
its work.

If you would like to merge the changes than I will open a new
issue, because I've added an additional minor fix.


Greetings,
Daniel

dan-t:
One additional thing, for functions racer seems to keep the `pub` modifier
but not for struct fields. The `pub` modifier for struct fields could be
also displayed in the popup menu.


Greetings,
Daniel

dan-t:
And seeing the type of struct fields would be also nice!

PR #197, opened 2015-04-13
Expand file names when setting environment variables in Emacs.

### LINKED ISSUE #195: Weird bogofile errors in the Emacs binding
Using Rust nightly 2015-04-12 on Ubuntu 14.04 and HEAD of racer, upon completing `use std::io::B` (the only contents of the file) from Emacs it doesn't work. On the command line it works fine. Tested with a clean emacs environment in only rust-mode and company enabled.

When I debug `racer--candidates` [(racer.el:56)] (https://github.com/phildawes/racer/blob/master/editors/racer.el#L56), while evaluating the let form of `lines` I get the following message from racer:

```
Result: ("PREFIX 13,14,B" "bogofile:1:15: 1:16 error: expected one of `::`, `;`, or `as`, found `<eof>`" 
"bogofile:1 use std::io::B" "                         ^" "bogofile:1:15: 1:16 error: expected one of `::`, `;`, or `as`, 
found `<eof>`" "bogofile:1 use std::io::B" "                         ^")
```

Newlines added to clarify the message. But if I invoke it on the command line it prints out a few lines of what apparently are correct results.

ane:
This ensures that any file name is canonicalized and made absolute.

If the file name isn't canonical, `setenv` will not expand its contents.

Fixes #195.

### DIFF
--- a/editors/racer.el
+++ b/editors/racer.el
@@ -52,7 +52,7 @@
   "Run the racer complete command and process the results."
   (setq racer-tmp-file-name (concat (buffer-file-name) ".racertmp"))
   (racer--write-tmp-file racer-tmp-file-name)
-  (setenv "RUST_SRC_PATH" racer-rust-src-path)
+  (setenv "RUST_SRC_PATH" (expand-file-name racer-rust-src-path))
     (let ((lines (process-lines racer-cmd
 				"complete"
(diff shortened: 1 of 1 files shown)


phildawes:
Nice! Thanks v much @ane

Issue #202, opened 2015-04-16
Racer build fails


rohitjoshi:
OS: OSX 10.10.3
    cargo 0.0.1-pre-nightly (84d6d2c 2015-03-31) (built 2015-03-31)
    rustc 1.0.0-beta (9854143cb 2015-04-02) (built 2015-04-02)
    git branch: master , checked out today

$cargo build
   Compiling rustc-serialize v0.3.12
   Compiling toml v0.1.20
   Compiling racer v0.0.1 (file:///projects/rust/racer)
src/racer/ast.rs:41:17: 41:37 error: type `&mut syntax::parse::parser::Parser<'_>` does not implement any method in scope named `parse_stmt_nopanic`
src/racer/ast.rs:41         match p.parse_stmt_nopanic() {
                                    ^~~~~~~~~~~~~~~~~~~~
src/racer/ast.rs:54:13: 54:18 error: mismatched types:
 expected `syntax::ast::Crate`,
    found `core::result::Result<_, _>`
(expected struct `syntax::ast::Crate`,
    found enum `core::result::Result`) [E0308]
src/racer/ast.rs:54             Ok(e) => Some(e),
                                ^~~~~
src/racer/ast.rs:55:13: 55:28 error: mismatched types:
 expected `syntax::ast::Crate`,
    found `core::result::Result<_, _>`
(expected struct `syntax::ast::Crate`,
    found enum `core::result::Result`) [E0308]
src/racer/ast.rs:55             Err(FatalError) => {
                                ^~~~~~~~~~~~~~~
src/racer/snippets.rs:40:17: 40:27 error: mismatched types:
 expected `syntax::ptr::P<syntax::ast::ImplItem>`,
    found `core::result::Result<_, _>`
(expected struct `syntax::ptr::P`,
    found enum `core::result::Result`) [E0308]
src/racer/snippets.rs:40                 Ok(method) => {
                                         ^~~~~~~~~~
src/racer/snippets.rs:55:17: 55:32 error: mismatched types:
 expected `syntax::ptr::P<syntax::ast::ImplItem>`,
    found `core::result::Result<_, _>`
(expected struct `syntax::ptr::P`,
    found enum `core::result::Result`) [E0308]
src/racer/snippets.rs:55                 Err(FatalError) => {
                                         ^~~~~~~~~~~~~~~
src/racer/snippets.rs:41:27: 41:38 error: the type of this value must be known in this context
src/racer/snippets.rs:41                     match method.node {
[...]

PR #205, opened 2015-04-20
Depend on `syntex_syntax` rather than `syntax`

tamird:
Probably won't solve #202 due to https://github.com/erickt/rust-syntex/issues/2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,1 +1,7 @@
 language: rust
+rust:
+  - nightly
+  - 1.0.0-beta
+matrix:
+  allow_failures:
+    - rust: 1.0.0-beta
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,7 +2,34 @@
 name = "racer"
 version = "0.0.1"
 dependencies = [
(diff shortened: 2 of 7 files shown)


tamird:
Yeah, this is green on nightly but not beta due to https://github.com/erickt/rust-syntex/issues/2

phildawes:
Hi @tamird,

Thanks very much for working on this. I'm thinking this should go in a separate 1.0 branch rather than master due to the increase in build times, what do you think?

Cheers,

Phil

tamird:
@phildawes yikes, the difference is build time is indeed surprising. I'm hesitant to put things in a separate branch, but maybe it's okay to not merge this until after 1.0 is out and `syntex` works for real.

erickt:
fyi, syntex v0.4.0 now compiles with beta :)

phildawes:
Oh nice work @erickt. In that case we should probably merge with master now that we have a real shot at a 1.0 compatible racer.

tamird:
Yeah I'm working on getting this building successfully on 1.0 now.

phildawes:
Merged!

PR #210, opened 2015-04-25
Update rustc-serialize to 0.3.13 for rust nightly

winding-lines:
This fixes the compilation with the latest nightly as of Apr 25 2015.

PR #211, opened 2015-04-26
Newer version of dependencies

iopq:
Fixes https://github.com/phildawes/racer/issues/207


phildawes:
Thanks @iopq! Closing because applied #210

PR #217, opened 2015-04-28
Work toward compiling on rust beta

tamird:
Reduces use of `#![feature()]` attributes.

@phildawes


phildawes:
Awesome, thanks @tamird

PR #222, opened 2015-05-02
Some code to make test compatible with rust beta and minor improvements.

damienrg:


golddranks:
I think Cargo.lock should'n be generally gitignored on end-user-executable builds. Cargo Guide mentions about that: http://doc.crates.io/guide.html#cargo.toml-vs-cargo.lock

damienrg:
Ok, it is also mentionned in the [faq of cargo] (http://doc.crates.io/faq.html#why-do-binaries-have-cargo.lock-in-version-control,-but-not-libraries?).
I will remove this commit and reorder other commits.

damienrg:
The commits are now updated please review the pull request.

phildawes:
Hi Damien,  Thanks for these. I like the regex patch, but txt_matches() is used a lot and I think it slows performance a bit.

E.g. on my laptop 
```rust
fn main() {
    let v = Vec::new();
    v.
}
```

On my laptop, before:

```
~/src/rust/racer$ time ~/src/rust/racer/target/release/racer complete 3 6 /tmp/a97.rs > /dev/null

real	0m0.338s
user	0m0.273s
sys	0m0.064s
```
after:
```
~/src/rust/racer$ time ~/src/rust/racer/target/release/racer complete 3 6 /tmp/a97.rs > /dev/null

real	0m0.422s
user	0m0.327s
sys	0m0.096s
```

I haven't used the rust regex library before. Have you tried the regex! macro?

damienrg:
regex_macro is unstable so for now we cannot use it. I will install rust nightly do to some benchmarks.
Will you apply some patches if i split this pull request in 3 pull requests:
- 1 for beta stuff
- 1 for fs::metadata
- 1 for txt_matches

damienrg:
I close this pull request and the new pull requests are:
1. for beta stuff #227 
2. for fs::metadata #228
3. wip

PR #224, opened 2015-05-03
Corrected find_ident_end for identifiers ending with a multibyte UTF-8 char

inrustwetrust:
There was an implicit assumption in find_ident_end that the last character of the identifier was encoded as a single UTF-8 byte. If this was not the case, the index returned would cause an "index does not lie on character boundary" panic when used.


phildawes:
Nice catch! Thanks v much @inrustwetrust

PR #225, opened 2015-05-03
Update travis to use 1.0.0-beta.4.

damienrg:
If pull request #222 is accepted could you also apply this as the code should work with rust beta and rust nightly.


damienrg:
Please look at #227 as it includes this commit.

PR #227, opened 2015-05-03
Make it possible to run test on rust beta and make bench only available with feature nightly.

damienrg:


phildawes:
Awesome, Thanks @damienrg

PR #228, opened 2015-05-03
Use std::fs::metadata to test if a path is a directory.

damienrg:


phildawes:
Thinks @damienrg 
Do you know if there are any problems with this on windows?

damienrg:
Unfortunately I cannot test it but a user on the rust channel confirms that racer is still working after this commit.

PR #229, opened 2015-05-03
Add test case for issue 223.

damienrg:
Note:
I tested on windows that before the fix the test failed and pass after the fix.


phildawes:
Thanks!

PR #230, opened 2015-05-03
Fix bench compilation error.

damienrg:
Add 'extern crate test' removed by mistake in 93954f697baa50df2f010da54acb9d825082d15c.


phildawes:
Thanks again

PR #235, opened 2015-05-12
Do not define '#' prefixed names outside autoload directory

rhysd:
Names prefixed with `#` are used for 'autoload' stuff.  For example, `foo#` prefix is assumed to be used in `autoload/foo.vim`.  Currently `racer#` prefix is used outside autoload directory in this repository, but it means that it may not work if `autoload/racer.vim` exists.

I removed `racer#` prefixes and simply add `Racer` prefix to define global functions.

Please see `:help autoload` for more detail

### DIFF
--- a/plugin/racer.vim
+++ b/plugin/racer.vim
@@ -37,7 +37,7 @@ if !exists('g:racer_insert_paren')
     let g:racer_insert_paren = 1
 endif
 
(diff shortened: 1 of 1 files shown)


phildawes:
Thanks v much @rhysd. Sorry for the delay in merging

rhysd:
Thanks for merge!

PR #239, opened 2015-05-17
Some improvements for the vim plugin

lilydjwg:
1. use `tempname()` to get temporary filenames
2. keep the position in jumplist before jumping away with 'gd' so the user can jump back with <C-o>

### DIFF
--- a/plugin/racer.vim
+++ b/plugin/racer.vim
@@ -134,6 +134,7 @@ function! racer#JumpToLocation(filename, linenum, colnum)
         if a:filename != bufname('%')
             exec 'e ' . fnameescape(a:filename)
         endif
+        normal! m`
         call cursor(a:linenum, a:colnum+1)
     endif
 endfunction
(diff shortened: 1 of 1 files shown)


phildawes:
Hi @lilydjwg, 
Thanks v much for this PR. I'm afraid I don't think the tempname() change will work, since racer currently uses the directory of the file to locate other local modules and crates so it needs to be written in the same directory as the original file. I agree that writing new files in the current directory isn't guaranteed to work - I think the best solution is one that allows passing the temporary file to racer via stdin or something, but that is obviously a bit more work.

In the meantime, maybe vim could only write temporary files if the file has been edited and otherwise use the existing file? that would cut down on the cases where writing new files wouldn't work. 

Thanks again, Phil

phildawes:
I applied the 'keep position in jumplist' patch, because I'm guessing that's useful to vim users. Thanks very much!

PR #248, opened 2015-05-22
vim plugin open definition in a split not buffer

bengesoff:
Seems nicer for me? Opens a vertical split instead of a new buffer for the definition after 'gd'

### DIFF
--- a/plugin/racer.vim
+++ b/plugin/racer.vim
@@ -135,7 +135,7 @@ endfunction
 function! RacerJumpToLocation(filename, linenum, colnum)
     if(a:filename != '')
         if a:filename != bufname('%')
-            exec 'e ' . fnameescape(a:filename)
+            exec 'vsp ' . fnameescape(a:filename)
         endif
         call cursor(a:linenum, a:colnum+1)
     endif


phildawes:
Hi @bengesoff, 
Thanks for the patch! I don't really use the vim plugin so I don't know if this is a desired feature or not. In emacs I prefer to jump the whole frame.
Could a regular vim user chime in here please? 

-Phil

tafia:
I think could be a nice feature. Would it be possible to have both ?  gd
and some other command ?

ches:
The behavior of Vim's standard `gd` function which the mapping in Racer's plugin shadows does not open a split, so I think changing the behavior in the way it's done in this patch is undesirable, replacing the current buffer is expected.

This can be achieved though by adding your own mappings. This works for a vertical split:

```vim
nnoremap <buffer> gd :vsplit<CR> :call RacerGoToDefinition()<CR>
```

Do that via an `autocmd FileType rust` or in `~/.vim/after/ftplugin/rust.vim`.

Personally I'm a fan of using the preview window which doesn't take my cursor out of the original file and can be dismissed consistently from anywhere with `<C-w>z`. For various filetypes with a specialized semantic go-to-definition plugin like Racer, I follow a convention of using `<LocalLeader>` and then mimicking standard Vim mappings for ctags jumping, e.g. `:h CTRL-W_]`:

```vim
nnoremap <buffer> <LocalLeader><C-w>] :split<CR> :call RacerGoToDefinition()<CR>

" No native Vim equivalents of these:
nnoremap <buffer> <LocalLeader><C-v>] :vsplit<CR> :call RacerGoToDefinition()<CR>
nnoremap <buffer> <LocalLeader><C-t>] :tab split<CR> :call RacerGoToDefinition()<CR>
```

Preview window equivalents (e.g. `<C-W>}`) are unfortunately more of a hassle because [Vim has a bug with losing cursor position][1] sometimes with the series of commands you'd naturally use to `:pedit` then move back to the original window. If you're determined to get them working, check how I've used ``m``` in a wrapper function [on this vim-erlang-tags patch][2].

If there's general interest I can send a PR for the Racer README showing a few suggested mappings that follow native analogs. I don't think actually adding them to the plugin proper is a good idea, because a prefix like `<LocalLeader>` is opinionated (though sensible for `<LocalLeader>`'s intent) and some people would likely not want un-prefixed ctags mappings overridden.

[1]: http://vim.1045645.n5.nabble.com/BUG-BufReadPre-autocmd-changes-cursor-position-on-pedit-td1206965.html
[2]: https://github.com/vim-erlang/vim-erlang-tags/pull/22

phildawes:
Thanks @ches, this is excellent information. A README PR would be most welcome.
I'm a vim n00b and this project gets a lot of questions and issues about the vim plugin. Better documentation would go a long way to reducing that

Issue #251, opened 2015-05-24
Prepare for MELPA


aminb:
We should be applying the suggestions in https://github.com/milkypostman/melpa/pull/2701#issuecomment-94887495 to have racer included in MELPA.

Wilfred:
Racer is now in MELPA, so this cna be closed.

phildawes:
Thanks @Wilfred

PR #252, opened 2015-05-25
Check some items off purcell's MELPA-compat list

tomjakubowski:
* Move the emacs package into its own directory (I think this will make
  future MELPA support easier)
* Add a Cask file for ease of development. See
  https://cask.readthedocs.org
* Add more standard headers to the package. See the "Library Headers"
  section of the elisp manual.
* Add package dependencies for emacs 24.3 and company 0.8.0.  I don't
  think racer actually needs to declare a cl-lib dependency, since that
  already ships with emacs 24.3, but I could be wrong!
* Fix a couple things that flycheck complained about.
* Rename the string/ends-with top level function to follow coding
  conventions.

Part of #251

### DIFF
--- a/editors/emacs/Cask
+++ b/editors/emacs/Cask
@@ -1,0 +1,4 @@
+(source gnu)
+(source melpa)
+
+(package-file "racer.el")
--- a/editors/emacs/racer.el
+++ b/editors/emacs/racer.el
@@ -1,8 +1,46 @@
 ;;; racer.el --- Rust completion via racer
 
+;; Copyright (c) 2014 Phil Dawes
+
+;; Author: Phil Dawes
+;; URL: https://github.com/phildawes/racer
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "24.3") (company "0.8.0") (rust-mode "0.2.0"))
+;; Keywords: abbrev, convenience, matching, rust, tools
+
+;; This file is not part of GNU Emacs.
+
+;; Permission is hereby granted, free of charge, to any
+;; person obtaining a copy of this software and associated
+;; documentation files (the "Software"), to deal in the
+;; Software without restriction, including without
+;; limitation the rights to use, copy, modify, merge,
+;; publish, distribute, sublicense, and/or sell copies of
+;; the Software, and to permit persons to whom the Software
+;; is furnished to do so, subject to the following
+;; conditions:
+
+;; The above copyright notice and this permission notice
+;; shall be included in all copies or substantial portions
+;; of the Software.
+
+;; THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
+;; ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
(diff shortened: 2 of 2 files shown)


tomjakubowski:
Issues I did not address from https://github.com/milkypostman/melpa/pull/2701#issuecomment-94887495:

> * It's not appropriate to automatically rebind TAB and M-. for the user: that should be the user's choice.

I agree with these concerns, but I don't think I'm the right person (read: familiar enough with racer) to make these changes. Pretty much that entire `add-hook` invocation seems questionable to me, though. 

> * Since this is a company plugin, it would arguably be better to call the package (and the file) company-racer.

That would a rather breaking change for me to make as a first contribution :-)

phildawes:
Thanks @tomjakubowski, this looks great.
Out of interest, why the test dependencies in the cask file? (ecukes etc..)

tomjakubowski:
Those were added by default when I used cask's CLI to make the cask file. I've removed them.

phildawes:
Great, thanks v much!

phildawes:
N.B. I also changed the README to reflect the new emacs directory. 
(this is a breaking change for existing emacs users so they'll need to update their .emacs files)

damienrg:
I think it is better to create a new github repository to only provide an emacs package and just keep the racer engine here.
There is already a company-racer repository https://github.com/emacs-pe/company-racer created by @emacs-pe and I think it will be better to work together on the emacs package.

tomjakubowski:
That sounds like a good idea to me @damienrg. One disadvantage of that is it requires close co-operation between the two repositories to be sure they're in sync with respect to command line arguments, output formats, etc.

phildawes:
Hi @tomjakubowski, @damienrg,

I agree with both of you - a separate project for racer emacs support with somebody else maintaining it would be a real help to me and much better for emacs users.

I'm not very familiar with github groups. A rust-racer group sounds like a good idea. My only concern is the human problem of nominating people and blessing 'official' repositories. I like github because of the decentralized structure. Is it possible to 'link' to existing repositories rather than transferring them?

I'll create the rust-racer group to try and get some experience with it. Thanks!

marsam:
Hi:
Sorry for the late response.  I think it's better to make `racer.el` independent from any autocompletion system.
There are features that `racer.el` could support (and I'd like to help): `imenu`, `eldoc`, and `go-to-definition` (à la `slime-describe-symbol`).

btw, I've wrote a bit about the [differences of `company-racer` and the current `racer.el`](https://github.com/emacs-pe/company-racer/issues/1#issuecomment-105739096), sorry for not making a pull request, for some reason `racer.el` didn't work for me, so I end up writing `company-racer`.

dleslie:
@phildawes @marsam 

FWIW, the outcomes of this discussion will likely invalidate #319.

dleslie:
Allow me to +1 @marsam; it seems to me that core Emacs functionality ought to be in ```racer.el```; like ```eldoc```, ```imenu```, ```go-to-definition```, and ```completion-at-point-fn```. Leave support for optional packages to the efforts of external maintainers, and save yourselves some work. :smile:

tomjakubowski:
I second everything @marsam and @dleslie said.

PR #258, opened 2015-05-31
Handle crate structure edge cases

Furyhunter:
This PR fixes two bugs which I ran into when using completions on the SDL2 and GL crates. Specifically, when a crate uses an additional root directory under `src` for its root module, racer would not correctly handle it, and would panic because `lib.rs` did not exist in the place it expected it to.


phildawes:
Great thanks @Furyhunter !

PR #262, opened 2015-06-03
Clean nameres

tafia:
Lot of changes to *nameres.rs*.
I have splitted all commits so I can probably revert the ones you don't like.

Main changes are:
- add constructors for `Match` and `PathSegment`
- use more iterator constructs, with the benefits:
  - do not check for `ExactMatch` for each item
  - use extend whenever possible
  - sometimes do not even create intermediary vector, just return the existing `vec::IntoIter`
  - most of the time, it is shorter. It looks easier to read for me

I haven't modified the biggest function `search_scope` because the alternatives I found didn't look good.

If needed I can squash it.

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -32,28 +32,12 @@ pub fn with_error_checking_parse<F, T>(s: String, f: F) -> Option<T> where F: Fn
 
 // parse a string, return a stmt
 pub fn string_to_stmt(source_str: String) -> Option<P<ast::Stmt>> {
-    with_error_checking_parse(source_str, |p| {
-        use syntex_syntax::diagnostic::FatalError;
-        match p.parse_stmt_nopanic() {
-            Ok(p) => p,
-            Err(FatalError) => None
-        }
-    })
+    with_error_checking_parse(source_str, |p| p.parse_stmt_nopanic().unwrap_or(None))
 }
 
 // parse a string, return a crate.
-pub fn string_to_crate (source_str: String) -> Option<ast::Crate> {
-    with_error_checking_parse(source_str.clone(), |p| {
-        use std::result::Result::{Ok, Err};
-        use syntex_syntax::diagnostic::FatalError;
-        match p.parse_crate_mod() {
-            Ok(e) => Some(e),
-            Err(FatalError) => {
-                debug!("unable to parse crate. Returning None |{}|", source_str);
-                None
-            }
-        }
-    })
+pub fn string_to_crate(source_str: String) -> Option<ast::Crate> {
+    with_error_checking_parse(source_str.clone(), |p| p.parse_crate_mod().ok())
 }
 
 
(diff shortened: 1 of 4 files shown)


phildawes:
Wow thanks @tafia! It's going to take me a while to go through these but from an initial glance through the commits these changes look good.

tafia:
No pressure !

tafia:
Just adding few more commits.
Note: i've removed [`from_svec`](https://github.com/tafia/racer/commit/143c74a026f9884fb6dc21efd01a4e5ff7727ea8#diff-8ace77072741b2439e0c5cdc28f2ac78L174) function as it can be replaced by the modified [`from_vec`](https://github.com/tafia/racer/commit/143c74a026f9884fb6dc21efd01a4e5ff7727ea8#diff-8ace77072741b2439e0c5cdc28f2ac78R157)

tafia:
I see 3 steps I'd like to work on to leverage more on Iterators (and be as lazy as possible)

1. use `Iterator` instead of `Vec` (*should be the simplest part*)
  - don't return `Vec` even in `matchers` (replace by iters)
  - remove `collect()` as much as possible (eg. return `FlatMap` instead of `vec::IntoIter`), it may not be possible for all fn but I already see several candidates
2. remove all the `SearchType`s (*moderately difficult*)
  - its main reason (I think) is to allow early returns if `ExactMatch`, but if 1. is accomplished, it won't be necessary anymore (a simple `next()` instead of `collect()` + perhaps a `find_ident_end` could be done once only at the very end (ie. in the `mod::search_expressions(..)`)
  - should remove lot of redundant `match search_type ...`
3. Have `Match`m `Ty` etc ... not owning the `Path`, sourcestr and searchstr
  - I *think* they are all immutable so a simple borrow may be enough
  - it'll add a viral `'a` everywhere ...
  - definitely harder to do ... but it is not linked to 1. nor 2.
  - I am not even sure it is possible, it just looks bad to clone everywhere

What do you think, am I going in a wrong direction ?
So far I plan to do all of it after this current clean, as it already tries to rely more on iterators

phildawes:
Re: lazyness and iterators: 

I attempted to do this some time ago with racer but failed at it pretty bad, hence the current Vec + SearchType thing.

I had something like python's 'yield' in mind, and attempted to do this with an iterator that wraps a lambda, and by chain()ing them together. The downsides were:

 - return types in rust must be fully typed, so they ended up massive and really brittle to changes.
   I worked around this by boxing the iterators into trait objects, but a boxed iterator doesn't support Iter, so I had to wrap that in its own iterator.
 - debugging the lazy code was significantly more difficult than eager code.

In the end I gave up. Don't let my failure dissuade you if you think you can make this work, but be mindful of complexity tradeoffs. 

Sorry I haven't got to reviewing your patches yet. Am hoping to get some time later today

Btw, thanks for your help on racer, I really appreciate it!

phildawes:
(n.b. also rust has changed a lot since I tried this)

tafia:
Closing it for now

PR #269, opened 2015-06-13
check for unsafe

### LINKED ISSUE #267: unsafe functions are not present in completion match lists
The clearest example of this is `racer complete std::str::from`, which should list `from_utf8` and `from_utf8_unchecked`. However, the latter is omitted. Public unsafe functions from other modules are similarly omitted from completion matches.

tafia:
close #267


phildawes:
Nice! Thanks @tafia

PR #270, opened 2015-06-13
search for all combinaisons of hyphen/underscore in crates

### LINKED ISSUE #226: Racer doesn't seem to find sources for crates with hyphens in their name
Running Racer in RustDT (http://rustdt.github.io/), it seems Racer doesn't give suggestions for "rustc-serialize", which is linked as `extern crate rustc_serialize` even though the crate name itself is `rustc-serialize`.

tafia:
It's a tentative attempt to close #226 
I couldn't make a test so I'm not sure it even works.
Any help is welcomed :smile:

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -41,21 +41,30 @@ pub fn string_to_stmt(source_str: String) -> Option<P<ast::Stmt>> {
     })
 }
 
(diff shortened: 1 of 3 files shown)


tafia:
doesn't work ...

PR #271, opened 2015-06-13
search for underscored names in lock files

### LINKED ISSUE #226: Racer doesn't seem to find sources for crates with hyphens in their name

tafia:
close #226

I have fixed the case where the crate is searched in Cargo.lock file, I haven't tried for Cargo.toml `find_src_via_tomlfile` as I didn't have crate examples to test.

I tested on the xsv project which uses rustc-serialize.

Again I couldn't do a simple test.

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -42,7 +42,7 @@ pub fn string_to_stmt(source_str: String) -> Option<P<ast::Stmt>> {
 }
 
 // parse a string, return a crate.
-pub fn string_to_crate (source_str: String) -> Option<ast::Crate> {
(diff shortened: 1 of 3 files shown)


tafia:
Pushed the fix on toml files as well.
- search for `cratename.replace("-", "_")`
- if `version == "*"`, then search only for the first matching folder

phildawes:
Looks good to me! Thanks v much @tafia

Issue #273, opened 2015-06-14
Racer uses a ridiculous amount of resources in some situations


nstoddard:
I use Sublime Text with Racer, on Windows 7, and I had started typing the following line when Racer started using a ton of resources:

```
panic!("trying
```

Racer started using the hard drive a lot (and possibly using other resources, I couldn't tell), and it froze my computer (I couldn't do anything but move the mouse cursor). I had to hold the power button to reboot. This happened while I was typing the word 'trying'. After I rebooted, I tried this again, and the same thing happened, but I was able to kill racer before my computer froze completely.

I don't know why this particular line triggered the problem; racer has worked fine on everything else so far.

phildawes:
Hi @nstoddard,
Thanks for the report. I suspect racer got itself into some sort of loop. I can't reproduce this with just the line 'panic!("trying', what else was in the file?

Thanks,

Phil

nstoddard:
I was working on a multiset data structure. I didn't save an exact copy of the code, but it looked something like this (I took the finished file and tried to remove stuff that I wrote after I encountered the problem):

```
use std::collections::hash_map::*;
use std::hash::Hash;
use std::ops::*;

pub struct Multiset<K: Eq + Hash> {
    elems: HashMap<K, u32>,
    // This is a stupid hack to avoid 'does not live long enough' in get()
    zero: u32,
}

impl<K: Eq + Hash> Multiset<K> {
    pub fn new() -> Self {
        Multiset{elems: HashMap::new(), zero: 0}
    }

    pub fn insert(&mut self, val: K) -> u32 {
        match self.elems.entry(val) {
            Entry::Vacant(entry) => {
                entry.insert(1);
                1
            }
            Entry::Occupied(mut entry) => {
                let view = entry.get_mut();
                *view += 1;
                *view
            }
        }
    }

    pub fn get(&self, val: K) -> &u32 {
        self.elems.get(&val).unwrap_or(&self.zero)
    }

    pub fn remove(&mut self, val: K) -> u32 {
        match self.elems.entry(val) {
            Entry::Vacant(_) => panic!("Unable to remove elem from Multiset (vacant)"),
            Entry::Occupied(mut entry) => {
                if *entry.get() == 0 {
                    panic!(   <--- I was writing this line
                }
            }
        }
    }
}
```
I was writing the line containing the `panic` near the bottom. This code was in 'main.rs' (I also had an empty main() function for testing) and I wasn't using Cargo.

[...]

PR #275, opened 2015-06-16
Fix searching enum definition in match arm

tafia:
When testing #273, I encountered an issue in `search_scope_headers`.
We can early return in case we are in the enum part of the match arm.

Added a test which doesn't work on current master

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,10 +24,11 @@ dependencies = [
 
 [[package]]
 name = "kernel32-sys"
-version = "0.1.0"
+version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


tafia:
Just see the commits, I can rebase on master so it'll look better if needed.

phildawes:
Hi @tafia. This looks good, thanks! If you could rebase that would be great

tafia:
sure, I'll do this asap

phildawes:
Hi @tafia. Thanks again, I rebased your fix onto master.

tafia:
Sorry !
I should have done it. Very sorry !

phildawes:
No worries, thanks for the fix!

PR #278, opened 2015-06-18
Make it possible to complete on temporary file

damienrg:
Most editors use a temporary file over the current file to send a
completion or find-definition request to racer. If this temporary
file is not in the project directory, the completion will not work
for external crates as racer searches a Cargo file in the parent 
directories of the file.

To fix this problem and to not break backward compatibility this commit
adds an optional argument to specify the file name which belongs to the
project.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -75,14 +75,19 @@ fn complete(match_found: &Fn(Match)) {
             }
             let charnum = args[3].parse::<usize>().unwrap();
             let fname = &args[4];
+            let substitute_file = Path::new(match args.len() > 5 {
+                true => &args[5],
+                false => fname
+            });
             let fpath = Path::new(fname);
-            let src = core::load_file(&fpath);
-            let line = &*getline(&fpath, linenum);
+            let src = core::load_file(&substitute_file);
+            let line = &*getline(&substitute_file, linenum);
             let (start, pos) = util::expand_ident(line, charnum);
             println!("PREFIX {},{},{}", start, pos, &line[start..pos]);
 
+            let session = core::Session::from_path(&fpath, &substitute_file);
             let point = scopes::coords_to_point(&*src, linenum, charnum);
-            for m in core::complete_from_file(&*src, &fpath, point) {
+            for m in core::complete_from_file(&*src, &fpath, point, &session) {
                 match_found(m);
             }
         }
@@ -96,7 +101,7 @@ fn complete(match_found: &Fn(Match)) {
                 if p.len() == 1 {
                     match_found(m);
                 } else {
-                    for m in do_external_search(&p[1..], &m.filepath, m.point, core::SearchType::StartsWith, core::Namespace::BothNamespaces) {
+                    for m in do_external_search(&p[1..], &m.filepath, m.point, core::SearchType::StartsWith, core::Namespace::BothNamespaces, &m.session) {
                         match_found(m);
                     }
                 }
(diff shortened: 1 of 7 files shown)


phildawes:
Hi @damienrg, Thanks for this - I agree this is important functionality. 

Unfortunately racer potentially reloads the file from the filename in a couple of places when recursing through the ast, so this patch won't work in some cases. (e.g. grep for load_file in the source)

I guess the solution is to either pass the file mapping around (e.g. in a context object or something), or put the mapping in TLS. I'm not sure which is best, but TLS is probably easier / less invasive.

damienrg:
My bad, I thought that the file was read only one time and that the content was borrowed all the time.
I might be wrong but it seems that racer does not use threading at all, so I thing that TLS should not be use in this case and I prefer the context object idea.

What solution do you prefer:
1. replace the Path argument with a context object containing Path of the project file and Path of the temporary file
2. add a context object as a new parameter
3. modify code to read the content only one time and always borrow the content (maybe it is not possible?)

phildawes:
Hi @damienrg.

I think the TLS approach equates to a global variable in the case where we don't switch threads. However I do think passing the mapping around is the better approach, it's just more work.

Replacing Path with a Pair of paths might not work: Rust code can have circular import dependencies, and so the original module (with the temporary file) could end up importing code from itself. When racer searches through another module we need to make sure the mapping is not lost in case it comes back to the original file.

Maybe adding a context object as a new parameter is the best way? That would also give a future way to do (3) easily

damienrg:
>Replacing Path with a Pair of paths might not work: Rust code can have circular import dependencies, and so the original module (with the temporary file) could end up importing code from itself. When racer searches through another module we need to make sure the mapping is not lost in case it comes back to the original file.

I am not sure I understand what you mean because of my lack of experience with racer internals.

>Maybe adding a context object as a new parameter is the best way? That would also give a future way to do (3) easily

I agree. I plan to add a FileContext or CompletionContext struct which could evolve to include more context if needed. Then, I plan to add this context as a new parameter to methods who need this.
What do you think about this?

phildawes:
That sounds like a good approach to me, thanks.

damienrg:
Weird, I am not able to reproduce the error. All tests pass even if i stress my pc.

marvelm:
Looks like it's only broken in the nightly. I can confirm that the tests pass locally for rustc 1.0.0.

phildawes:
Hi @damienrg.

Thanks for this. Sorry I didn't explain myself very well: 
I was expecting the context object to contain a mapping rather than a path,
e.g.  "/path/to/mymodule.rs" -> "/tmp/tmpfile3boecb.rs"
..so that whenever racer tries to load 'mymodule.rs' it knows to instead load the temporary file in /tmp/tmpfile3boecb.rs. 

I think racer needs this because it might try and reload 'mymodule.rs' during its search, even though the original query was from that file.

damienrg:
>@marvelm 
Looks like it's only broken in the nightly. I can confirm that the tests pass locally for rustc 1.0.0.

I also tested on nightly and all tests pass.

>@phildawes
Thanks for this. Sorry I didn't explain myself very well:
I was expecting the context object to contain a mapping rather than a path,
e.g. "/path/to/mymodule.rs" -> "/tmp/tmpfile3boecb.rs"
..so that whenever racer tries to load 'mymodule.rs' it knows to instead load the temporary file in /tmp/tmpfile3boecb.rs. 

 Ok, so you mean that the context object should contain a map (or maybe there is a better type)?

>@phildawes
I think racer needs this because it might try and reload 'mymodule.rs' during its search, even though the original query was from that file. 

From what I saw in the code:
- the content of the file is read, in this case we need to use the temporary file
- the path of the file is used to search crates, in this case we need to use the path of the original file

So you mean that I should use the original file all the time and use the temporary only when racer do load_file..., right?
If this is right, my solution (which handles theses cases) is not as clean as your solution so I will update the pull request.

phildawes:
> So you mean that I should use the original file all the time and use the temporary only when racer do load_file..., right?
Yes exactly! Sorry I wasn't clear earlier

damienrg:
@phildawes can you give me the declaration of the struct FileContext that you have in mind?

phildawes:
I guess it just needs to remember the original path and the temporary file, so something like
```rust
struct Session {
    query_path: &PathBuf,            // the input path of the query
    substitute_file: &PathBuf        // the temparory file
}
```
will do for now. (I prefer Session to Context now I think about it, though I'm not very good at naming things!). As long as it supports the requirements you outlined I think it'll be fine.
I guess over time racer might add more stuff (like cached source files). Also at some point we might make load() a method on session.

damienrg:
Please review this patch.
This change was error prone so I may have introduce a bug even if I have added some assert_eq! and had done some tests without hitting a problem.
If you prefer I can update this pull request to remove the assert_eq! calls.

phildawes:
Hi @damienrg, this looks awesome. I'll spend some more time testing today/tomorrow and then will merge. 
Thanks again for all your work on this.

PR #283, opened 2015-06-21
Refactor emacs lisp for MELPA release

Wilfred:
Hi Phil

I've made changes to racer.el so it should be ready for release on MELPA (see #251). I've also refactored a few little bits.

Do let me know what you think, I'd love to see this package available for download from MELPA.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@
 
 3. Install company mode. (e.g. via ELPA: ```M-x list-packages```, select ```'company'```)
 
-4. Set some variables and install racer. E.g. add this to .emacs:
+4. Configure Emacs to find racer:
 
    ```
    (setq racer-rust-src-path "<path-to-rust-srcdir>/src/")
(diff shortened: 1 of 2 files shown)


phildawes:
Hi @Wilfred,
This looks great. The only thing is to make sure the setup docs in the README work for emacs beginners (otherwise it generates issue/support overhead). 
When I tried from scratch I noticed the following lines don't work on startup in a .emacs because rust-mode is autoloaded. 
```
(define-key rust-mode-map (kbd "TAB") #'racer-complete-or-indent)
(define-key rust-mode-map (kbd "M-.") #'racer-find-definition)
```
I'm no emacs guru, what do you think is the best solution? Should we recommend people add them to a rust-mode-hook? 
Thanks very much!

Wilfred:
Aha, well spotted! This should say:

```lisp
(require 'rust-mode)
(define-key rust-mode-map (kbd "TAB") #'racer-complete-or-indent)
(define-key rust-mode-map (kbd "M-.") #'racer-find-definition)
```

I mentioned the `require` in the header in the racer.el file, but missed it in the README.md. I'll push a fix.

Wilfred:
@phildawes this should now be fixed :)

phildawes:
Awesome, thanks @Wilfred

PR #284, opened 2015-06-25
check CARGO_HOME environment variable when looking for Cargo root dir

kirillkh:


phildawes:
Nice! Thanks @kirillkh

PR #290, opened 2015-06-28
Further elisp fixes

Wilfred:
Hi Phil

A few additional improvements to racer.el.

1. Fix a crash in `racer--complete-at-point-fn`. I introduced this in my last PR and didn't notice because it doesn't crash if you had loaded the old racer.el in your current emacs session. Sorry about that.
2. Improve package description, as suggested by the MELPA team.
3. Minor corrections to the cask file.

### DIFF
--- a/editors/emacs/Cask
+++ b/editors/emacs/Cask
@@ -2,3 +2,6 @@
 (source melpa)
 
 (package-file "racer.el")
+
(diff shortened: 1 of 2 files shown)


phildawes:
Cool, thanks @Wilfred

PR #291, opened 2015-06-28
Add coverage support

damienrg:
This is based on https://users.rust-lang.org/t/tutorial-how-to-collect-test-coverages-for-rust-project/650.


damienrg:
Close for now as it seem that coveralls does not parse the json file correctly.

Issue #287, opened 2015-06-27
Default racer command


Yamakaky:
The default value for `racer_cmd` is relative to the vim plugin. You should also check for `racer` alone, if it is in the PATH. You could also add the target/ dir to the PATH.

phildawes:
Sounds reasonable - patches gratefully accepted!

Shougo:
I think it should be closed.
It is already implemented.

PR #293, opened 2015-07-01
Add better defaults for racer_cmd and RUST_SRC_PATH

Yamakaky:
See #287


phildawes:
Thanks @Yamakaky

Yamakaky:
The improved RUST_SRC_PATH could also be used for the binary. Do you want me to try ?

PR #294, opened 2015-07-01
Emacs: Better racer-cmd default

moonlite:
Make `racer-cmd` in racer.el default to whatever `racer` is available in
the system PATH.

FIXES: #294


moonlite:
Pushed the PR by mistake, meant to:
1) attach it to #294
and
2) test it before sending.

But yeah. Will test in a bit if you don't beat me to it.

moonlite:
Yep it seems to work fine for me! :)

phildawes:
Great, thanks @moonlite

PR #295, opened 2015-07-02
Use silent :write

Shougo:
Current write messages are really annoying.
I fixed it.
Please merge it.


phildawes:
Nice. Thanks @Shougo

Shougo:
Thank you!

PR #300, opened 2015-07-05
Ignore file which cannot be open

damienrg:
We read files returned by do_file_search and as directory may contain
files which cannot be opened this commit remove them from the result.

Emacs may create a lock file when a buffer is modified and not saved:
http://www.gnu.org/software/emacs/manual/html_node/emacs/Interlocking.html
as this file cannot be read it make racer panics when trying to read the file.

Note: I wanted to add a unit test but i do not know how i could create an invalid file.

### DIFF
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -367,7 +367,7 @@ pub fn do_file_search(searchstr: &str, currentdir: &Path) -> vec::IntoIter<Match
                         }
                         {
                             // try just <name>.rs
-                            if fname.ends_with(".rs") {
+                            if fname.ends_with(".rs") && File::open(&fpath_buf).is_ok() {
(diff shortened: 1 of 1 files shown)


phildawes:
Hi @damienrg,

Thanks for the patch! Does this affect performance much? - it looks like it now opens every file rather than just the ones matching the searchstring

damienrg:
I can do some benchmarks but it will depend on my disk speed and it will be
worse in case of a folder over network. So as you mention it will be better
if i delay the test. I will update the pull request.

damienrg:
@phildawes I have updated the patch to only open file when necessary. I have also done a small refactoring and if you prefer I can split it in 2 pull requests.

phildawes:
Great, thanks @damienrg

PR #303, opened 2015-07-10
Use the tag ring before jumping to a definition.

Wilfred:
Hi Phil

I've made a change to `racer-find-definition` so that it just works with `M-,` (which is `pop-tag-mark`). This is consistent with other emacs tools that jump to a definition (tags, slime-nav). What do you think?


phildawes:
Nice, thanks v much @Wilfred

PR #307, opened 2015-07-17
Make Vim plugin work on a buffer with no file name

chris-morgan:
This also stops the 'alternate-file' from being scrubbed.

As noted, this is not perfect; if I am merely ignorant of the proper fix then please feel free to fix it or tell me!


phildawes:
Thanks @chris-morgan. Merged! (I'm afraid I am also ignorant, sorry).

Issue #313, opened 2015-07-29
No support for auto-complete-mode


dleslie:
I've made this issue as this question is likely to arise.

Those wishing to use auto-complete-mode can probably make it work with ac-company.

https://raw.githubusercontent.com/buzztaiki/auto-complete/master/ac-company.el

syohex:
NOTE: I had written [ac-racer](https://github.com/syohex/emacs-ac-racer).

PR #319, opened 2015-08-02
Update racer.el

marsam:
Hi:

I've updated the `racer.el` package, to use `deferred.el` and separate the company-package to use `racer.el`. I think this change will simplify things like adding support for auto-complete (#313)

If this get merged I think the melpa recipe of `company-racer` can be updated to use this package.

Thanks in advance.

### DIFF
--- a/editors/emacs/Cask
+++ b/editors/emacs/Cask
@@ -3,5 +3,4 @@
 
 (package-file "racer.el")
 
-(depends-on "rust-mode")
 (depends-on "company")
--- a/editors/emacs/Makefile
+++ b/editors/emacs/Makefile
@@ -1,0 +1,29 @@
+CASK  ?= cask
+WGET  ?= wget
+EMACS  = emacs
+
+EMACSFLAGS =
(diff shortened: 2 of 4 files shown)


dleslie:
It would be nice if ```racer--complete-at-point-fn`` remained in some capacity, so that those not using ``company-mode`` specifically can tie it into Emacs' built-in ``completion-at-point-functions``.

In fact, IMHO Racer ought to tie into ``completion-at-point-functions`` by default, with ``company`` or other as optional extensions.

marsam:
I totally agree about `racer--complete-at-point-fn`. I'll update the pull request this weekend

dleslie:
Cool! Thank-you

dleslie:
Any news @marsam? :smile:

dleslie:
Given the discussion in #252 perhaps it is appropriate that this pull request be abandoned?

marsam:
Hi, I sorry for the late response, I was really busy these days :(. I'll close this pull request for now and follow the discussion.

PR #329, opened 2015-08-09
Handle win32 search path separator & add win32 src path

dwink:
When new defaults for `g:racer_cmd` and `$RUST_SRC_PATH` were added to the Vim plugin recently, the code assumed a colon as the path separator -- which is fine on all sane OS's. Alas, on Windows the separator is a semicolon. This left the $PATH variable broken if g:racer_cmd was not set when loading racer. This change detects `has("win32")` and does the right thing.

Also, added a reasonable default for the RUST_SRC_PATH on Windows while I was at it.


phildawes:
Thanks @dwink. Sorry for delay in merging

PR #333, opened 2015-08-15
Eldoc support

Wilfred:
Hi Phil

I've added eldoc support to racer (see the screenshot). I've also updated the screenshot and simplified the install instructions. Let me know if anything doesn't make sense or if you'd like changes :)


dleslie:
Ooh!

phildawes:
This is awesome, thanks @Wilfred .

phildawes:
I'm not an emacs expert by any stretch. I tried to follow the instructions on a fresh emacs install and hit
``` Symbol's value as variable is void: package-archives ```

I think I needed a  ```(require package)``` before it.

Also the first time I tried to ```M-x package-install RET racer RET``` It couldn't find the racer package, presumably because I had some cached elpa state somewhere or something. ```M-x list-packages``` sorted it. Is there a command for reloading package state?

**Comment by phildawes** [comment]
Also I get  ```File mode specification error: (void-function racer-turn-on-eldoc)```

I think this is because my emacs loaded ' racer Package-Version: 20150804.1258', which doesn't contain the eldoc. Does that sound right?

**Comment by phildawes** [comment]
I updated the README.md to  add the ```(require 'package)``` and remove eldoc (until the eldoc version hits melba. Hope that's ok

**Comment by phildawes** [comment]
He @Wilfred, I've just noticed I use the mark in order to get back to my previous position after a find-definition. What should I be doing instead? (thanks!)

**Comment by Wilfred** [comment]
Hi Phil

Thanks for keeping me honest, you're absolutely right that I was sloppy in my testing of the instructions :)

> I think I needed a (require 'package) before it.

Yep, that's correct.

> M-x list-packages sorted it. Is there a command for reloading package state? 

There's

**Comment by Wilfred** [comment]
> Hey @Wilfred, I've just noticed I use the mark in order to get back to my previous position after a find-definition. What should I be doing instead? (thanks!)

You just need to pop the tag mark.

```lisp
(require 'rust-mode)
(define-key rust-mode-map (kbd "M-,") #'pop-tag-mark)
```

We should probably add this to the documentation.

purcell:
> @purcell is there any easier way to get notified when MELPA has built a new version?

There's an existing issue I created about wanting to put maintainer twitter accounts in the recipes, and then /cc them when tweeting about builds. Would that do the trick?

Wilfred:
I think that's a nice idea. Does it scale to larger projects with multiple maintainers? For projects with longer descriptions the tweets look pretty long already.

We'd also need to be careful when creating recipes. Most of the MELPA recipes I've contributed are for projects created by others (I don't know if I'm representative). If I copied an example recipe without understanding this feature, I could set `:maintainer "jrandom"` and inadvertently spam the maintainer.

Wilfred:
Aha, I've found the relevant MELPA ticket: https://github.com/milkypostman/melpa/issues/2106

PR #341, opened 2015-08-18
Add location support, improve annotations, hilight eldocs when completing, enable automatic completions on certain prefixes (:: and .).

ane:
The completion now begins automatically (customizable) if point is preceded by `::` or `.`, defauls to `t`.

Annotations now appear on the first completion (there was some weird thing that caused
them to only appear after some interaction had occurred).

The company eldoc strings are now syntax highlighted.

Location jumping: hitting <kbd>C-w</kbd> on a candidate takes you to the file/line of the match.
Column jump should be possible but company does not support that.

The setting of `company-idle-delay` to `nil` was removed: the company user should decide
if the triggering is manual or not (so this is a company customization, not a racer one).


Wilfred:
Fantastic! Thanks for this :)

PR #344, opened 2015-08-20
Fix #330 use tempname() instead of racertmp

### LINKED ISSUE #330: Make a Vim variable that allows users to set the path of the tmp file
I would like to be able to override the path of where the racertmp files get written.

E.g. Instead of the current directory, a user could point to `/tmp` as the default

Shougo:
I fixed #330 problem.
Please check it.

### DIFF
--- a/plugin/racer.vim
+++ b/plugin/racer.vim
@@ -57,7 +57,7 @@ function! RacerGetPrefixCol()
     let fname = expand("%:p")
     let tmpfname = tempname()
     call writefile(getline(1, '$'), tmpfname)
-    let cmd = g:racer_cmd." prefix ".line(".")." ".col." ".tmpfname
+    let cmd = g:racer_cmd." prefix ".line(".")." ".col." ".fname." ".tmpfname
     let res = system(cmd)
     let prefixline = split(res, "\\n")[0]
     let startcol = split(prefixline[7:], ",")[0]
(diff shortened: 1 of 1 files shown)


phildawes:
@Shougo, this looks awesome. I'll give it a quick run-through tonight or tomorrow and then merge.  Thanks!

Shougo:
:-)

phildawes:
Hi @shougo,

Sorry for the delay in reviewing the patch. I'm afraid I don't think this works as is: The main reason the existing plugin writes the temp file in the same directory is so that racer can find the other modules, Cargo.toml etc.. from the file position. Unfortunately after this patch racer fails to find other local modules because it searches as if they are in the /tmp directory.

E.g. Open racer/src/bin/main.rs in vim and then try to 'gd' on a function reference. (e.g. go to line 20, col 25, hit 'gd' to go to 'do_file_search'). 

Luckily racer has recently added functionality to support tempfiles in a different directory: The patch needs to pass the original filename to racer, but then can pass the temporary filename as an additional arg.

My vim is pretty rubbish, but I guess this will be:

```
     let cmd = g:racer_cmd." complete ".line(".")." ".col." ".fname." ".tmpfname
```

Hope that makes sense,

Phil

Shougo:
@phildawes Thank you.  Fixed.

phildawes:
I gave it a quick test, looks good. Thanks!

Shougo:
Thank you!

PR #350, opened 2015-08-23
Emacs cleanup

### LINKED ISSUE #339: Don't remove other company modes
The emacs mode sets the `company-backends` variable to be just `(racer-company-complete)`:
```
(set (make-local-variable 'company-backends) '(racer-company-complete))
```

It should instead add this function to the list, so the other modes aren't overwritten. I find `company-dabbrev-code` to be a useful backend, and it's annoying to have to renable it with racer installed.

fbergroth:
I ended up doing quite a lot of changes, see commits for more details.

Setting up `racer-mode` should be more straightforward now:
```el
(add-hook 'rust-mode-hook #'racer-mode)
(add-hook 'racer-mode-hook #'eldoc-mode)
(setq company-tooltip-align-annotations t) ; looks better
```
I also removed the dependency of company since completion-at-point is supported out of the box via `company-capf`.
Fixes #339.

Cheers :)

### DIFF
--- a/editors/emacs/racer.el
+++ b/editors/emacs/racer.el
@@ -75,17 +75,13 @@
   :group 'rust-mode)
 
 (defcustom racer-cmd
-  (if (locate-file "racer" exec-path)
-      (locate-file "racer" exec-path)
-    "/usr/local/bin/racer")
+  (or (executable-find "racer") "/usr/local/bin/racer")
   "Path to the racer binary."
   :type 'file
   :group 'racer)
 
 (defcustom racer-rust-src-path
-  (if (getenv "RUST_SRC_PATH")
-      (getenv "RUST_SRC_PATH")
-    "/usr/local/src/rust/src")
+  (or (getenv "RUST_SRC_PATH") "/usr/local/src/rust/src")
   "Path to the rust source tree."
   :type 'file
   :group 'racer)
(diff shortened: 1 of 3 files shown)


phildawes:
Hi @fbergroth. Thanks for this patch, really appreciated.
I haven't had a proper chance to test yet, but am merging so it will get picked up by melba and will test then.
Cheers, Phil

editors/emacs/racer.el
```diff
@@ -296,27 +184,30 @@ foo(bar, |baz); -> foo|(bar, baz);"
...
+
+(define-minor-mode racer-mode
+  "Minor mode for racer."
+  :lighter " racer"
+  :keymap racer-mode-map
+  (setq-local eldoc-documentation-function #'racer-eldoc)
+  (make-local-variable 'completion-at-point-functions)
+  (add-to-list 'completion-at-point-functions #'racer-complete-at-point))
+
+(define-obsolete-function-alias 'racer-turn-on-eldoc 'eldoc-mode)
```
Wilfred: Shouldn't this be `(define-obsolete-function-alias 'racer-turn-on-eldoc 'racer--eldoc)`?

fbergroth: `racer-eldoc` is called by eldoc, and not by the user. `racer-turn-on-eldoc` used to set the `eldoc-documentation-function` (which is now done by `racer-mode`) and activate `eldoc-mode`. I believe this is correct.

fbergroth: Cider does it similarly in [`cider-eldoc.el`](https://github.com/clojure-emacs/cider/blob/0c5b393b545806a48d30c441fdfe31274e5e7949/cider-eldoc.el#L157)

Wilfred: Aha, makes total sense. (I've learnt a number of new things reading your patch :) ).

PR #351, opened 2015-08-24
Emacs: Error if no definition is found

fbergroth:
Better than being silent.

### DIFF
--- a/editors/emacs/racer.el
+++ b/editors/emacs/racer.el
@@ -136,18 +136,19 @@
 (defun racer-find-definition ()
   "Run the racer find-definition command and process the results."
   (interactive)
-  (-when-let (match (--first (s-starts-with? "MATCH" it)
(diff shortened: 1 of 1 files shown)


syohex:
I suppose `message` is better than `error` in such case.

fbergroth:
`error` seems to be more common, such as in [elpy](https://github.com/jorgenschaefer/elpy/blob/ea71a12822207391010ab02de778808179b42e41/elpy.el#L1594) and [elisp-slime-nav](https://github.com/purcell/elisp-slime-nav/blob/f6d241b11abbc4064e17e02dbd0bc6c61db256cc/elisp-slime-nav.el#L100). I guess it's easier to handle if you want to call `racer-find-definition` from elisp.

syohex:
I see

phildawes:
Thanks @fbergroth

Issue #353, opened 2015-08-24
company: duplicated information in completion menu


birkenfeld:
Since the merge of #350, when `company-tooltip-align-annotations` is nil, the element names and annotations are smashed against each other without any spacing. Also, the old annotation of "Type : Signature" was better than the new "Signature : Type".

Finally, this is not new, but the whole function signature is *very* long to display in the menu. I would suggest to

* remove the element name from the annotation as this is just duplicated information
* truncate after a certain length

cc @fbergroth 

![shot1](https://cloud.githubusercontent.com/assets/144359/9447208/76ace7e6-4a96-11e5-8ee1-43e3dc8bac6e.png)
![shot2](https://cloud.githubusercontent.com/assets/144359/9447209/76b058cc-4a96-11e5-9c8a-f2411f8023e6.png)

fbergroth:
I agree, it's ugly with `company-tooltip-align-annotations` set to `nil`.

Trimming off the name is a good idea. With trimming, wouldn't it look better to have the completions outlined as "Signature : Type", like:
```
ConstantFloat (String, FloatWidth) : EnumVariant
ConstantInt (u64, LitIntType) : EnumVariant
```
Then it almost matches the actual signatures.

I'm just starting out with rust, so I might be missing some case.

birkenfeld:
> With trimming, wouldn't it look better to have the completions outlined as "Signature : Type"

I agree.

birkenfeld:
BTW, I also wondered why in the second screenshot some enum variants don't have a signature. I traced that to a bug in `s.el`:

https://github.com/magnars/s.el/pull/75

If that doesn't get fixed in due time, you might want to consider a workaround.

birkenfeld:
ok nvm ;)

fbergroth:
Nice catch :)

PR #356, opened 2015-08-24
Enhance emacs completions

fbergroth:
As discussed in #353. Try it out @birkenfeld. Truncation still needed?


birkenfeld:
Looks good. Still can get long lines, but I think it's definitely an improvement. Thanks!

phildawes:
Thanks again!

PR #360, opened 2015-08-25
nameres: search only in function signature for "fn prefix" (fixes #359)

### LINKED ISSUE #359: inner functions are suggested with prefix
When calling `complete` on `struct.` with a prefix, inner functions are suggested that aren't suggested when there is no prefix.

Example file:
```rust
struct T(u32);

impl T {
    fn outer(&self, x: u32) {
        fn inner(y: u32) { }
    }
    fn index(&self) { }
}

fn main() {
    let t = T(0);
    t.i
}
```

Calling racer:
```
$ racer complete 12 6 test_file.rs         # point after "t."
PREFIX 6,6,
MATCH 0,1,9,test_file.rs,StructField,0
MATCH outer,4,7,test_file.rs,Function,fn outer(&self, x: u32)
MATCH index,7,7,test_file.rs,Function,fn index(&self)
END
$ racer complete 12 7 test_file.rs         # point after "t.i"
PREFIX 6,7,i
MATCH inner,5,11,test_file.rs,Function,fn outer(&self, x: u32)
MATCH index,7,7,test_file.rs,Function,fn index(&self)
END
```

birkenfeld:


phildawes:
Great, thanks for the fix @birkenfeld

PR #361, opened 2015-08-25
Fix #206 add g:loaded_racer variable and compatible option check

### LINKED ISSUE #206: Add loaded_racer variable for selective loading / avoiding reloading and feature testing in racer.vim
A pattern a lot of vim plugins follow is having a loaded variable that halts processing of the script if it has already been loaded, or if under some conditions the user wants it not to be loaded they can set the variable prior to script init. 

```viml
" This goes at the very top of racer.vim
if exists ("g:loaded_racer")
    finish
endif
let g:loaded_racer=1
```

It is common enough for required feature checks to also go on the if line: 

```viml
if exists("g:loaded_racer") || &cp || v:version < 700 
```

Shougo:
I implemented the feature in #206.

### DIFF
--- a/plugin/racer.vim
+++ b/plugin/racer.vim
@@ -8,6 +8,14 @@
 "
 " (This plugin is best used with the 'hidden' option enabled so that switching buffers doesn't force you to save) 
 
+if exists('g:loaded_racer')
+  finish
+endif
+
+let g:loaded_racer = 1
+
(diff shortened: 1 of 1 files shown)


phildawes:
Cool, thanks @Shougo

Shougo:
Thanks.

PR #362, opened 2015-08-25
[RFC] Clippy lint fixes

birkenfeld:
This is the result from running rust-clippy over the racer source.

It's probably best to review the commits individually.

Since a lot of "idiomatic rust" lints can be overruled by personal preference, please let me know which of these changes are not desired in the racer codebase.

There are also objective improvements, such as using `to_owned` which is faster for string conversion, and one instance of using `into_iter` to avoid string clones.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -83,12 +83,12 @@ fn complete(match_found: &Fn(Match), args: &[String]) {
             let fpath = Path::new(fname);
             let session = core::Session::from_path(&fpath, &substitute_file);
             let src = core::load_file(&fpath, &session);
-            let line = &*getline(&substitute_file, linenum, &session);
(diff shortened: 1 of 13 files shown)


tafia:
looks good :+1:

phildawes:
Thanks for taking the time to do this @birkenfeld, is really helpful and much appreciated.

Most of these are where the compiler now elides things that used to have to be made explicit. It really shows how the language has become less noisy! 

I noticed sometimes to_string() got changed to to_owned(), and sometimes to '.into()'. Do you know the reason?

birkenfeld:
Thanks for the merge!

As for the string methods, for literals I prefer `into()` since it's shorter, but otherwise `to_owned()` is somewhat more explicit.

PR #364, opened 2015-08-26
Cache file contents in the session(s)

birkenfeld:
This series of commits makes an attempt to avoid re-reading files from disk over and over. It also makes all Session access use a `Rc` pointer, so that the session isn't cloned for every match.

This gives a large speedup in debug mode (doing `racer complete std::io::` drops from 1.14s to 0.48s on my machine), probably because the `BufReader` implementation is really slow when unoptimized. 

In a release build I get a ~25% speedup from 30ms to 22ms using kbknapp's method from here: https://gist.github.com/kbknapp/fb28a49d7a0fe8b8e400 - allocations are down from from 82k (19M bytes) to 56k (7M bytes).

In daemon mode every request creates its own root Session, so files are not cached. This is probably a good thing as long as the cache is never invalidated e.g. based on mtime.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -82,7 +82,7 @@ fn complete(match_found: &Fn(Match), args: &[String]) {
             });
             let fpath = Path::new(fname);
             let session = core::Session::from_path(&fpath, &substitute_file);
-            let src = core::load_file(&fpath, &session);
+            let src = session.load_file(&fpath);
             let line = &getline(&substitute_file, linenum, &session);
             let (start, pos) = util::expand_ident(line, charnum);
             println!("PREFIX {},{},{}", start, pos, &line[start..pos]);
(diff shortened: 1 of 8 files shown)


birkenfeld:
(updated benchmark numbers in description)

dleslie:
As a person who codes on a slower machine, allow me to say thank-you.

phildawes:
Awesome, thanks @birkenfeld! Have merged.

BTW, I might be wrong but I don't think the session needs to be returned in the match object, I think it was just convenience when @damienrg added the Session to handle temporary file translation. 

If instead we passed the session as a parameter to all calls then we wouldn't need RC/RefCell because mutable borrows would suffice. Does that sound correct to you?

Cheers!

birkenfeld:
@phildawes so there is basically no reason to have multiple Session objects? If so, that can simplify the matter a lot.

birkenfeld:
We'll likely still need the RefCell for the file cache, since using a mutable borrow is going to be very unpleasant - many of the functions make liberal use of multiple borrowing.

PR #366, opened 2015-08-26
codeiter and codecleaner fixes/speedups

birkenfeld:
(This is based on the file_cache branch.)

A few fixes and speedups for the codeiter and codecleaner iterators, which are shown to be hotspots by oprofile. Doesn't give nearly as much speedup as the file caching, though.


phildawes:
Thanks for the patch. I probably won't get a chance to look/merge this until tonight, sorry!

phildawes:
This is great, thanks!

PR #369, opened 2015-08-27
Remove session references in Match and other objects

birkenfeld:
As discussed in #364. Session is now passed explicitly to all functions that need it. `Match`, `Scope` and `PathSearch` don't have a session field anymore. As a result, we can drop the `Rc` around the session reference.

Interestingly, that means that a few functions gained the `session` argument, but a few others also lost it, because they only used the session to compare a match.session.filepath with another filepath in an `assert`.

You'll be happy to hear that this also brings a speedup of 1.1ms/21.8ms = 5% for the usual benchmark case.


phildawes:
Nice! Thanks again

PR #370, opened 2015-08-27
ast: speed up ParseSess construction with terminal stderr

birkenfeld:
If racer runs with stderr going to a terminal this change saves a lot of time (>40%) reading terminfo files to determine color support, which is entirely unnecessary for racer.

This is probably mostly be relevant for people testing or benchmarking racer, since in normal use it will probably run with stderr going to a pipe or redirected to `/dev/null`.


phildawes:
Cool thanks

PR #371, opened 2015-08-28
create list of code indices immediately when loading files

birkenfeld:
The `codecleaner` iterator yields indices that don't change as long as this source lives. Now that we cache the source, we can also calculate these indices only once and cache them together with the source.

(Unfortunately this is not easily extended to `iter_stmts` because the results from there depend on which point the iteration starts. A more complicated data structure would be needed for that.)

Added a dependency on the `typed-arena` crate. The arena allows us to cache strings/IndexedSource objects and still get long lived references out of the RefCells.

This gives a ~33% speedup in the usual benchmark, 9.7 to 6.6 ms.


phildawes:
Looks good, thanks v much

PR #373, opened 2015-08-28
nameres: fix matching of trait-impl methods (fixes #349)

### LINKED ISSUE #349: Racer doesn't discover String::from
It discovers the other `String::from_...` methods, so I'm not sure what's different about `String::from`.

birkenfeld:
Methods in trait impls are never declared "pub", but still visible
outside their module.


phildawes:
Oh cool. Thanks for the fix

PR #378, opened 2015-08-30
Implement name lookup for more loop headers

birkenfeld:
This implements name (and type) lookup for more loop headers, i.e. `for-in` and `while let`.


phildawes:
Awesome! Thanks @birkenfeld

PR #380, opened 2015-08-30
nameres: fix misrecognition of "impl X for Y"

birkenfeld:
for and while loops only can occur at stmt start anyway.


phildawes:
Nice! Thanks @birkenfeld

PR #384, opened 2015-09-01
fixes for #381 (const functions)

### LINKED ISSUE #381: racer completes some static functions weirdly
```
% racer complete std::cell::RefCell::
MATCH fn new(value,308,14,/Users/tjakubowski/src/rustc-1.2.0/src/libcore/cell.rs,Const,pub const fn new(value: T) -> RefCell<T> {
MATCH default,472,7,/Users/tjakubowski/src/rustc-1.2.0/src/libcore/cell.rs,Function,fn default() -> RefCell<T> {
```

That `"fn new(value"` gets inserted literally into the completion in Emacs:

screen shot 2015-08-30 at 6 49 06 pm
screen shot 2015-08-30 at 6 49 09 pm

On the other hand, `default` and `String::new()` are completed just fine.

jaxx:
Trying to fix #381. Digging through some Rust source I found out that there can be functions with `const` keyword also. Racer thinks that this is a constant instead of a function and this messes up the parsing. Maybe this is a hack, but you can reject this if it is :)

### DIFF
--- a/src/racer/matchers.rs
+++ b/src/racer/matchers.rs
@@ -93,6 +93,10 @@ fn find_keyword(src: &str, pattern: &str, search: &str, search_type: SearchType,
     }
 }
 
+fn is_const_fn(src: &str, blobstart: usize, blobend: usize) -> bool {
+    src[blobstart..blobend].contains("const fn")
+}
+
 fn match_pattern_start(src: &str, blobstart: usize, blobend: usize,
                        searchstr: &str, filepath: &Path, search_type: SearchType,
                        local: bool, pattern: &str, mtype: MatchType) -> Option<Match> {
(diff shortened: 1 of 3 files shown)


phildawes:
Hi @jaxx. This looks great, sorry I didn't get to it earlier. I'll have a quick check through and then merge

phildawes:
Hi @jaxx 
Thanks again for the patch. It did occur to me that any function that contains an inner const function won't complete after this patch e.g.
```rust
#![feature(const_fn)]
fn blah() {
    const fn foo() {
    }
}

fn main() {
    b     <---- doesn't complete to blah
}
```

But that's not as big a deal as #381 so have merged. Thanks again!

PR #385, opened 2015-09-01
Updated installation instructions

christetreault:
Installation instructions previously started at the step where you already have the source code. Updated to add step 1: get source from git.


phildawes:
Thanks @christetreault

PR #391, opened 2015-09-04
codecleaner: fix char literal handling

birkenfeld:
A single quote begins a char literal if either
* the next character is a backslash
* or the character after next is a single quote again


phildawes:
Cool thanks. According to travis this broke the 'follows_use' test, but I couldn't reproduce this failure. Maybe we've got a sporadically failing test :(

birkenfeld:
I'll try to investigate. FTR, I almost always run the tests locally before submitting a PR too, and didn't see a spurious failure here.

Issue #342, opened 2015-08-19
vim: Allow opening definition in new split-window


marcianx:
It would be nice to have, eg, gD open the definition in a new split-window, after which one could use gd to explore further within that window.

phildawes:
I don't use vim personally, but I'd be open to merging such a patch unless somebody else disagrees.

jwilm:
@phildawes looks like this has been resolved

MattWis:
Can confirm, this should have been closed when my PR was merged in

PR #397, opened 2015-09-10
Vim: Add #342, gD acts like gd in new vsplit

MattWis:
This adds the feature of being able to use gD to open the definition in a new vertical split, as requested in #342.


phildawes:
Thanks @MattWis, this is cool.
(I gave the patch a quick test and then merged)

PR #417, opened 2015-10-28
Remove deoplete source

Shougo:
Hi, I split deoplete source from racer.
It should be integrated with vim-racer.

https://github.com/racer-rust/vim-racer/pull/9


phildawes:
Thanks @Shougo

Shougo:
Thanks!

Can you merge https://github.com/racer-rust/vim-racer/pull/9 ?

PR #425, opened 2015-11-14
Add defaults to RUST_SRC_PATH

Yamakaky:
If RUST_SRC_PATH is not set, use `/usr/src/rust/src` or `/usr/local/src/rust/src`.


Dr-Emann:
/usr/local usually should override /usr, it should check /usr/local first.

Yamakaky:
Done

phildawes:
Thanks @Yamakaky

PR #426, opened 2015-11-16
Dependency updates and bugfix

pmarcelll:


phildawes:
Cool, thanks @pmarcelll

Issue #326, opened 2015-08-07
Use an argument parsing library


marvelm:
Would you be opposed to me adding another dependency (https://github.com/tailhook/rust-argparse) and cleaning up the argument handling in main.rs ?

phildawes:
Sounds good. Is it the code you want to clean up, or were you thinking of
changing the arguments? If the latter we will need to make sure plugins
remain compatible

marvelm:
Sorry for not responding sooner.
I tried to use rust-argparse, but it didn't seem to improve the code quality. I think we should use Docopt (https://github.com/docopt/docopt.rs) instead. It takes a usage string, a struct, and decodes args into the struct. I have an example of this at https://github.com/marvelm/racer/blob/effe804/src/bin/main.rs#L188
I don't think the commands/arguments will change.

cburgdorf:
I suggest using github.com/kbknapp/clap-rs. It has the best API I've seen so far and is very well maintained by @kbknapp

kbknapp:
Thanks for the kind words! I've put together a gist of what `main.rs` would look like using `clap` at https://gist.github.com/kbknapp/f71a574d89abbe966d8d

This is truly a one-for-one port, all functionality should be exactly the same, plus all the extras added by `clap` for free, so to speak (Noted below).

A few things to note about this:

* There are some oddities when using the first argument after `complete` as *either* a line number or a fully-qualified-name. This one-for-one port says, "The first argument is required (either line number or FQN), and only if a second argument is found (`charnum`) then a third argument also becomes required (`fname`)" There is probably a cleaner way to do this, but not without changing the interface in some manner (such as adding flags, or changing subcommand names, etc.)
* You get suggestions for free; say a user types `racer compete std::io::B` (note the missing `L` in comp**l**ete) they would see an error saying, "the subcommand 'compete' isn't valid, did you mean 'complete'?"
* You get "smart usage strings" or what I sometimes call "context aware" usage strings. For example, if you were to run `racer complete 250 20` you'll get an error saying `fname` wasn't provided but is required, in the suggested usage string, the user will see `racer complete <fqn or linenum> [charnum] [fname]` even though `charnum` and `fname` are technically optional, they aren't in this *context* so they are included.
* I work pretty hard to keep `clap` efficient, so you get all the above with *very, very* few allocations.

I'd recommend trying it out, playing with some errors, testing speed / efficiency, etc. If you'd like, I can put in a PR.

[...]

PR #427, opened 2015-11-20
uses clap to parse command line arguments

kbknapp:
A possible implementation for #326 which doesn't sacrifice any performance, but gains a good bit in the command line user-friendliness department...at least IMO :wink: This PR would also make it  easier to add functionality to the command line in the future (such as possibly dealing with multiple rust sources as [mentioned recently](https://www.reddit.com/r/rust/comments/3thorf/multirust_and_racervim_dont_play_nice_together/) in [/r/rust](https://reddit.com/r/rust))

Something to note, this PR uses a slightly hacky way to stay 100% compatible with previous `racer` command line binaries. Specifically that one could do either:

```
$ racer complete <fqn>
$ racer complete <linenum> <charnum> <fn_name> [substitute_file]
```

The first argument to the `complete` command has a dual purpose. If 100% backwards compatibility isn't desired, there may be a more clean way to solve this. Having said that, the way this PR implements it works just fine, and to the user, there is no difference from previous releases.

A final note, I have run some simple tests, using this branch as my main `racer` for personal development for a little bit to work out any bugs, but I don't know how much of the full functionality I've actually tested. So if there are formal tests (beyond `cargo test`) I'd recommend running them against this branch just to check.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,6 +2,7 @@
 name = "racer"
 version = "1.0.0"
 dependencies = [
+ "clap 1.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "syntex_syntax 0.19.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


phildawes:
This looks great, thanks @kbknapp

phildawes:
Hi @kbknapp ,
I'm afraid I had to revert. It looks like find-definition no longer works, sorry:

```
target/release/racer find-definition 2 14 /tmp/a.rs
error: The argument 'fqn' was not found or is not valid

USAGE:
	racer find-definition [FLAGS] <linenum> <charnum> <fname> [ARGS]

Please re-run with --help for more information
```

kbknapp:
Ah OK no worries! Let me see if I can find a good test for that and update.

PR #430, opened 2015-11-23
fixes find-definition subcommand

kbknapp:
This fixes #427. I tested the output of each subcommand against the output from master here an all matched up :smile:


phildawes:
Looks good, thanks @kbknapp!

PR #436, opened 2015-11-28
Fix Offsets of CRLF files

### LINKED ISSUE #424: Newer versions of racer seem broken, including v1.0.0 tag
So, I had an old executable of racer, and it's working fine (I don't remember when I compiled it, nor which commit it was based on).

With a source like this:
```
struct Circle {
    x: f64, y: f64,
}

impl Circle {
	fn blah() {
	}
}

pub fn fooA() {
	f
}

fn fooB() {
	C
}
```

If I ask for completions after the "f", I get flate, fmt_macros, fooA, fooB. Invoking after "C", I get Circle, Clone, Copy. Seems good.

Now, if try to build racer from master, or from the v1.0.0 tag, it is broken. If I ask for completions after the "f", I get fooA, and the same happens if I request invocation after "C".

I'm on Windows, Rust 1.5.

### LINKED ISSUE #434: Completion of `use` suddenly fails
Hello! When I try to complete the following
```
use std::
```
(`racer complete 1 9 test.rs`) - all is well, but completion of
```
use std::env;
use std::
```
(`racer complete 2 9 test.rs`) unrespectively fails.

### LINKED ISSUE #435: Racer complete can't handle CRLF source files
If the source file is encoded with CRLF instead of LF, it miscounts the offset when using ``racer complete <linenum> <charnum> <path> [substitute_file]`` in  [core::complete_from_file](https://github.com/phildawes/racer/blob/2a6352926782274c430f3fe619dbf3ef375c56f3/src/racer/core.rs#L421) and therefore searches for the wrong completion.

Update: Turns out the function is already called with the wrong pos. It's always the same, no matter your line endings, which is obviously wrong.

CryZe:
``coords_to_point`` and ``point_to_coords_from_file`` assume that the line endings are ending with a simple ``LF`` and not a ``CRLF``, which is why they get the offset wrong which then causes Racer to complete the wrong parts of the code. By using ``split('\n')`` instead of ``lines()``, the code can still assume that the ``LF`` is missing, but the ``CR`` is still part of the individual splits, so it gets counted properly.

Fixes #435, Fixes #434, Fixes #424. I suspect this might fix some of the other issues people are having on Windows too.

### DIFF
--- a/src/racer/scopes.rs
+++ b/src/racer/scopes.rs
@@ -269,7 +269,7 @@ pub fn end_of_next_scope(src: &str) -> &str {
 
 pub fn coords_to_point(src: &str, mut linenum: usize, col: usize) -> usize {
     let mut point = 0;
-    for line in src.lines() {
+    for line in src.split('\n') {
         linenum -= 1;
         if linenum == 0 { break }
         point += line.len() + 1;  // +1 for the \n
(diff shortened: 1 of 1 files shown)


phildawes:
Awe. Some. Thanks @CryZe!

Issue #438, opened 2015-12-01
Does not complete all methods of boxed objects (excludes inner type)


Jonesey13:
Completion for methods of a `Box<T>` object only gives completions for `Box` and not for `T`. For example: 

```
let boxed_string = Box::new(String::from_str("Hello")));
boxed_string.  /// Complete here
```
Completing `boxed_string.` would not find the `len()` method from the type `String`.

PR #439, opened 2015-12-01
Extended completion of methods for boxed types

Jonesey13:
Resolves issue #438 by adding an additional enum variant to `Ty` and extending some match blocks in the `ast` and `nameres` modules.


phildawes:
This looks great @Jonesey13, thanks very much!

PR #444, opened 2015-12-11
Mention ST RustAutoComplete plugin as Sublime Text integration

nabijaczleweli:


phildawes:
Thansk @nabijaczleweli

PR #445, opened 2015-12-11
"tabbed text" interface mode

vadimcn:
The current Racer's interface is not well suited for programmatic interaction:
- Outputs are comma/semicolon separated, however in some cases output may contain names of generic Rust types, which may contain commas.  This makes parsing of Racer output ambiguous.
- Input commands in `deamon` mode are whitespace-separated and contain file paths, which is again ambiguous, since paths may contain spaces.

This change introduces a new "tabbed text" interface mode, in which all separators are replaced with tabs (i.e. `\t`), which will allow unambiguous data parsing in both directions.
This applies to both input and output (however does not affect command line parsing, as that is done by the OS).

For compatibility with existing clients, "tabbed text" mode is activated only when Racer is launched with the `--interface=tab-text` option.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -27,35 +27,58 @@ use std::path::{Path, PathBuf};
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
 
 #[cfg(not(test))]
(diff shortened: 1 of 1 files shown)


phildawes:
Thanks @vadimcn, merged.

Issue #422, opened 2015-11-09
Q: racer integration, find_definition usage


jwilm:
I recently started [integrating racer for YouCompleteMe](https://github.com/jwilm/racerd) support, and this is the first of what I'm sure will be many questions :smile:. This question regards `find_definition`.

```rust
pub fn find_definition(src: &str, filepath: &path::Path, pos: usize, session: SessionRef) -> Option<Match>
```

One of the arguments is a source buffer. If I don’t write the buffer to disk first, I get a panic (full bt below).

	thread 'engine::racer::tests::find_definition' panicked at 'index 0 and/or 89 in `` do not lie on character boundary', ../src/libcore/str/mod.rs:1441

However, writing the buffer to disk completely resolves that problem. It seems to me that, given a buffer, reading the initial file should be unnecessary. Is this a requirement? I would like to change the behavior to not do that if possible. I actually couldn’t find where it was doing that IO, either. Additionally, since this method is doing I/O, maybe it should return a Result instead of panicking while indexing a slice.

### Backtrace
```
   6:        0x10c490097 - rust_begin_unwind
   7:        0x10c4b4060 - panicking::panic_fmt::h6c78ce0128588a957HK
   8:        0x10c4b7b60 - str::slice_error_fail::hae4b5de9c67475548OS
   9:        0x10c0f44b5 - str::traits::_<impl>::index::he8945502e3410953mBS
  10:        0x10c0f437e - string::_<impl>::index::h547be55ef8b5db17i3f
  11:        0x10c0f3b92 - core::_<impl>::deref::h6728862e5c44cb4atMa
  12:        0x10c1030aa - scopes::mask_comments::h758821d339eb152bJeb
  13:        0x10c10e321 - scopes::scope_start::h49adfd7fdda43b7544a
  14:        0x10c1c6f90 - nameres::search_local_scopes::h2aa638ee8218b5856xe
  15:        0x10c1bc129 - nameres::resolve_name::ha8fef8b2d1cc0115pPe
  16:        0x10c10912e - nameres::resolve_path::hd3f93ff20151730eeZe
  17:        0x10c10c842 - core::find_definition_::h8b9c5b4772ebed00tYa
  18:        0x10c10c16b - core::find_definition::h36205eee38345da54Xa
  19:        0x10c0dc75b - engine::racer::_<impl>::find_definition::h6f5d2878fe1b3851yea
  20:        0x10c0e34c0 - engine::racer::tests::find_definition::h607071473acdcea1yla
[...]

PR #446, opened 2015-12-12
Add possibility to inject string into cache instead of file from disk.

bjornarg:
Adds the possibility to insert a string as the contents of a file in the cache, instead of reading it from a file from disk. Allows completion from in-memory strings without going through the filesystem.

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -414,6 +414,12 @@ impl<'s> Session<'s> {
             self.cache.arena.alloc(src.src.with_src(scopes::mask_comments(src)))
         }).as_ref()
     }
+
+    pub fn insert_file(&'s self, filepath: &path::Path, src: String) {
+        let mut cache = self.cache.raw_map.borrow_mut();
+        let index_source = IndexedSource::new(src);
+        cache.insert(filepath.to_path_buf(), self.cache.arena.alloc(index_source));
+    }
 }
 
 


phildawes:
Hi @bjornarg 
Thanks for the patch, do you have code that uses/requires this?

jwilm:
[racerd](https://github.com/jwilm/racerd) requires the ability to cache buffers like this. There are a few gotchas worth mentioning. First, it is important to also update the _masked_ version of the source as well. If the masked version had been accessed before updating the cache, future operations could result in a panic due to differences in the raw vs masked (imagine asking for a completion from a position in raw which doesn't exist in masked). Second, without changes to the `FileCache`, old allocations will be leaked. This is probably not an issue for short lived processes, but it is a critical issue for racerd.

[This branch](https://github.com/phildawes/racer/compare/master...jwilm:wip-reusable-cache#diff-032e27e14256ac8b3f95301690ae6a4dR418) attempts to address the mentioned issues. I hope to get this merged eventually, but there is currently some unsafe code in the `FileCache` implementation that needs to get refactored into an `Arena` implementation that supports allocation reuse (and thus remove the unsafe code from Racer). Further discussion of the racerd branch should probably be in #422 as to not hijack this thread.

phildawes:
Closing because functionality is in #452, thanks @bjornarg

PR #450, opened 2015-12-20
Add support for custom git branches.

bjornarg:
Checks the source string of any git repositories to see if a branch is
specified, since cargo uses it in the path.


phildawes:
Cool! thanks @bjornarg

PR #452, opened 2015-12-24
Add support for caching in-memory buffers to FileCache

jwilm:
**TL;DR**
- Adds `cache_file_contents` method to session for caching in-memory buffers/files
- Adds some unsafe code to `FileCache` / `Session` to support Arena allocation reuse
- API changes - `Session::from_path` takes additional argument, and `SessionRef` was obsoleted.

## Changes

The Session type now has an additional method `cache_file_contents` which enables consumers to write in-memory buffers into the file cache for use in racer operations. This feature is critical for projects like [racerd][] which run for extended periods and rely on caching for performance.

Initially, this introduced a memory leak since Arena allocations could not be reused, but this was fixed with some changes to the FileCache. The FileCache now proxies Arena `alloc` calls and first consults a free list. The `alloc` proxy was able to be implemented as a safe API, but the free operation `update_available_allocations` is inherently unsafe, and this is exposed as part of the API. To be 100% safe, this operation would need to know statically that there are no outstanding references to the item being freed. To guarantee this, file cache references are only accessed via a Session. When a session is dropped, it is known that it has no outstanding references, and at this point we can call `update_available_allocations`.

Since the Session now requires a reference to the FileCache, a bunch of the session lifetime restrictions were lifted. This accounts for many of the scattered changes in this PR including obsoleting the SessionRef type (thanks @birkenfeld). Additionally, all of the tests had to be updated to reflect the new Session API.

## Discussion

The caching feature added here is important for racer's story as a library. I think the main focus of discussion here should be with regards to how the Arena allocation reuse was achieved. Other proposed implementations of reap allocators use smart pointers to track active references and achieve a safe API. It's not clear to me whether the smart pointers would actually be any better than the previous `Rc` strategy the Session used. I would like to land this PR soon to ship [racerd][] for YCM using the official racer repository, but I understand if you would rather wait until an actual Reap allocator implementation is available.

[racerd]: https://github.com/jwilm/racerd

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -122,7 +122,8 @@ fn run_the_complete_fn(cfg: &Config, print_type: CompletePrinter) {
     let fn_path = &*cfg.fn_name.as_ref().unwrap();
     let substitute_file = cfg.substitute_file.as_ref().unwrap_or(fn_path);
 
-    let session = core::Session::from_path(fn_path, substitute_file);
+    let cache = core::FileCache::new();
+    let session = core::Session::from_path(&cache, fn_path, substitute_file);
     let src = session.load_file(fn_path);
     let line = &getline(substitute_file, cfg.linenum, &session);
     let (start, pos) = util::expand_ident(line, cfg.charnum);
@@ -149,7 +150,8 @@ fn external_complete(cfg: Config) {
     // input: a command line string passed in
     let p: Vec<&str> = cfg.fqn.as_ref().unwrap().split("::").collect();
     let cwd = Path::new(".");
-    let session = core::Session::from_path(&cwd, &cwd);
+    let cache = core::FileCache::new();
+    let session = core::Session::from_path(&cache, &cwd, &cwd);
 
     for m in do_file_search(p[0], &Path::new(".")) {
         if p.len() == 1 {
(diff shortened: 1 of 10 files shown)


phildawes:
Hi @jwilm. This looks good to me on first glance. I'd like to spend a little more time looking over this before I merge, so will get back to you hopefully later today. Thanks v much!

jwilm:
Thanks @phildawes! Looking forward to your feedback :smile:.

phildawes:
Hi @jwilm 
I'm afraid this currently doesn't work when the racer temporary file differs from the real file.
e.g. to reproduce
`$ echo "use std::" > /tmp/tmpfile`
then
`$ ./target/release/racer 1 9 /tmp/realfile.rs /tmp/tmpfile`
(i.e. realfile.rs doesn't exist yet)

I'm guessing this isn't a problem for racerd because the ycm protocol allows updating of unsaved files, but unfortunately other editor plugins rely on this to work with unsaved files.

Hope that makes sense. If you're busy I might get a chance to look and fix this later, but will probably be a bit busy today with family boxing day stuff. Thanks v much!

birkenfeld:
I have a slightly different idea that *should* require no unsafe code. I'll try to attempt an implementation soon, to see if the borrow checker agrees with me on that point :)

jwilm:
@phildawes Substitute file stuff is fixed, and a test was added.

@birkenfeld I'm excited to see what you come up with!

phildawes:
@jwilm, Awesome thanks, merged.
@birkenfeld: I'm also v interested

jwilm:
@phildawes thanks for merging this. Functionality in #446 is superseded by these changes; that PR should be safe to close.

Are you open to releasing an update on crates.io? I would love to put a semver dependency for racerd's racer dependency instead of git :grin:.

phildawes:
@jwilm I released v1.1.0 to github and crates.io. Hope I did it correctly!

jwilm:
Looks great. Thank you!

birkenfeld:
Sorry, nothing done yet. But I haven't forgotten :)

PR #456, opened 2015-12-30
Make cargo install work on windows

forki:
Otherwise it complains about invalid url


phildawes:
Thanks for pointing this out. Actually now that racer 1.1.0 has been released to crates.io, `cargo install racer` should be enough. I updated the README to reflect this.

Issue #455, opened 2015-12-27
Does not complete for inner type on lock guards.


yohanesu75:
Hi! Thank you for your effort on this wonderful project. It's a lifesaver for me :)

I would love it if it was possible to add completion for T when you have things like RwLockReadGuard<T>, RwLockWriteGuard<T>, MutexGuard<T>, etc.

Thank you.

Jonesey13:
I did this for Box<T> earlier but the approach I used was rather crude. This could easily be extended to the types you mentioned but doing this separately for each one involves a lot of repetition.

What would be ideal is to find implementations of Deref<T> for a given type and then look for methods on the type T.

PR #458, opened 2016-01-02
Completions for Deref Coercions

Jonesey13:
This fixes Issue #455 by recursively completing on coerced types (see the new tests for examples).   

Since this also retroactively solves issue #438 I've removed the changes from my earlier commit #439 as it is now redundant.

N.B. For the tests I had to copy over the definition of the Deref trait since it appears that travis.ci cannot find the source code for crate std.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,6 +1,6 @@
 [root]
 name = "racer"
-version = "1.0.0"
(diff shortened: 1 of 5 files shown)


phildawes:
Oh nice! Thanks v much @Jonesey13

PR #463, opened 2016-01-08
Added macro support

lfabreges:
Add support for macros


jwilm:
This appears to add support for locating macro definitions via find_definition. Looking forward to seeing this land. :smile:

phildawes:
This patch appears to match macros in the same source file. Merged, thanks @ifraixedes!
(n.b. In order to match macros like println!() racer will need to follow 'macro_use' directives I think)

PR #467, opened 2016-01-10
Read substitute file from stdin

henriiik:
Hello!

Not all editors use temp-files (or don't expose them to plugin authors). I made it possible for the substitute file to be read from stdin if you specify "-" as the substitute_file.

Is this functionality something you would be interested in merging? In that case i will continue to work on it. There are a few things that still needs doing.

- Check/fix for daemon mode
- Make testable and add tests

If you are curious, I am working on a plugin for [Visual Studio Code](https://code.visualstudio.com/).

The plugin is located at https://github.com/henriiik/vscode-rust.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -24,6 +24,8 @@ use racer::scopes;
 #[cfg(not(test))]
 use std::path::{Path, PathBuf};
 #[cfg(not(test))]
+use std::io::{self, BufRead};
+#[cfg(not(test))]
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
 
 #[cfg(not(test))]
@@ -118,12 +120,28 @@ enum CompletePrinter {
 }
 
 #[cfg(not(test))]
+fn cache_file_contents_from_stdin<'a>(file: &PathBuf, cache: &'a core::FileCache<'a>) {
+    let stdin = io::stdin();
+
+    let mut rawbytes = Vec::new();
+    stdin.lock().read_until(0x04, &mut rawbytes).unwrap();
+
+    let buf = String::from_utf8(rawbytes).unwrap();
+    cache.cache_file_contents(file, buf);
+}
+
+#[cfg(not(test))]
 fn run_the_complete_fn(cfg: &Config, print_type: CompletePrinter) {
     let fn_path = &*cfg.fn_name.as_ref().unwrap();
     let substitute_file = cfg.substitute_file.as_ref().unwrap_or(fn_path);
 
     let cache = core::FileCache::new();
     let session = core::Session::from_path(&cache, fn_path, substitute_file);
+
+    if substitute_file.to_str() == Some("-") {
+        cache_file_contents_from_stdin(&substitute_file, &cache);
+    }
+
     let src = session.load_file(fn_path);
     let line = &getline(substitute_file, cfg.linenum, &session);
     let (start, pos) = util::expand_ident(line, cfg.charnum);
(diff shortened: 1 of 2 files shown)


phildawes:
Hi @henriiik. This looks cool thanks. Were you thinking of coding an equivalent for daemon mode as well? I suspect over time racer will be used less in cmd and more as a long lived daemon. Also checkout https://github.com/jwilm/racerd, which provides a way to pass temporary files over http

Re vscode, I think @saviorisdead is also working on vscode support here: https://github.com/saviorisdead/RustyCode

henriiik:
If you are interested in this functionality (which you seem to be) I will gladly implement it for daemon mode as well. :)

Maybe i should have a look at adding something similar to the library API as well? It seems that racerd now has to save the POST'ed contents to disk.

I have used and contributed to RustyCode, but i found that it did "too much" for my tastes and i wanted to try and do something that suit me better. :)

jwilm:
@henriiik I highly recommend racerd for your use case. It was built to make integrating racer very easy for editors and IDEs (it powers [YouCompleteMe](https://github.com/valloric/youcompleteme)'s Rust integration). Performance has been a top priority, and there is definitely no file writing in any of its completion/goto functions. If you decide to use it, I am happy to provide whatever support you need. Just ping me somewhere on github :).

henriiik:
Thank you for your offer @jwilm! I only had a quick look at racerd on my phone this morning, hence my hasty comment. :)

I updated the PR earlier today with a fix for daemon mode and it seems to be working well in my initial tests.

My extension to vscode is mostly for my own use and learning/entertainment, but thought this would be a useful contribution to racer.

But is using the racer CLI not recommended and my time better spent elsewhere?

phildawes:
Hi @henriiik. At the moment using the CLI is fine (although not as performant as racerd). I use it all the time via emacs.
As racer tries to use rustc for type inference I suspect it will require a longer running process to perform acceptably (because rustc isn't designed for lazy parsing and analysis, so we may need to cache more).

henriiik:
Thnks for your reply @phildawes, I understand. The CLI has a daemon mode as well, is that not as performant as racerd either?

phildawes:
@jwilm has put a lot of effort into racerd and it has an emphasis on performance, so I would guess that if you need a daemon then racerd is probably your best bet.

To be honest, it makes sense to cut down on the number of ways to interface with racer in order to reduce maintenance overhead, so it might make sense for us to deprecate the racer cli daemon mode in favour of the racerd code. I need to have a chat to @jwilm and maybe the plugin authors about the best way to proceed with this

henriiik:
That sounds like a good plan @phildawes. :)

henriiik:
Hello again!

I have updated the code and squashed all the commits and the pull-request is now ready for merging unless you have any comments, which I will gladly listen to. Or if you think the changes are not necessary, feel free to close it. :) 

The code for caching from stdin now lives with the CLI in src/bin/main.rs, and if you specify "-" as the substitute file, it will read from stdin until it hits an EOF bit.

Here are some screenshot's of the changes in action:

normal mode
daemon mode

jwilm:
@phildawes I was hoping to discuss the future of racer and the racer ecosystem with you at some point. I'm working the rest of today, but I can make a post in the morning (GMT-8) outlining my thoughts.

bruno-medeiros:
@jwilm Yeah, what is the difference between Racer deamon mode and racerd ? Is it just the communication API, or does racerd provide more useful information to IDEs? 
Also, why is racerd using HTTP ? Seems like a unnecessary overhead, where just sockets + JSON messages should suffice...

jwilm:
@bruno-medeiros 

`racerd` has optimizations that aren't available in racer's daemon mode.
1. racerd keeps the file cache around between requests. This gives a big performance boost when completing from dependencies and especially the standard library. Granted, this would be easy to add to daemon mode in racer.
2. racerd supports any number of dirty buffers to use in a request. Aside from removing the need for temporary files completely, this allows completions to work in files the user hasn't saved.

The overhead of HTTP for localhost is effectively zero, and every practical language has some way to  make HTTP requests (making interop easy). Additionally, it is really easy to structure servers to use HTTP. For example, racerd response time for completing the line `use std::` has been been benchmarked at 7ms. This is *well* beyond the human threshold for instantaneous. Honestly, why not use HTTP?

phildawes:
@jwilm I'm also really snowed under with family and work at the moment. Sorry if it appears I'm ignoring you (or anybody else)

jwilm:
@phildawes Take your time! We are all volunteers here. Work/family should definitely take priority. :smile:

phildawes:
Thanks @henriiik, this looks great. I should have merged this much earlier than I did, sorry!

henriiik:
No problem @phildawes ! Thanks for the merge :)

PR #468, opened 2016-01-11
[NOT READY] Complete on separate lines

jwilm:
We received [an issue on ycmd](https://github.com/Valloric/ycmd/issues/301) that completions were not working when a trigger `.` started on the next line (example below). I traced this back to racer not supporting the feature.

So far, this PR is simply a failing test with the desired behavior. I have a couple of questions before starting work on a fix.
1. Does racer support completions for chained method calls? It would have to know about the return type of previous methods in order to perform the completion, and I'm not sure racer does that yet.
2. What would an appropriate fix for this be? I can imagine a super naïve solution where we eat up all whitespace characters preceding a trigger `.`. One issue with this I see immediately is needing to write a false representation of the file into the FileCache.
## Example

``` rust
struct Foo {
    pub barbaz: u32
}

fn main() {
    let thing = Foo { barbaz: 10 };
    // completion here should be barbaz, but this panics instead
    thing
         .
}
```

[Review on Reviewable](https://reviewable.io/reviews/phildawes/racer/468)

### DIFF
--- a/tests/system.rs
+++ b/tests/system.rs
@@ -47,6 +47,43 @@ fn completes_fn() {
 }
 
 #[test]
+fn completes_members_across_lines() {
+    let src="
+    struct Foo {
+        pub barbaz: u32
+    }
+
+    fn main() {
+        let thing = Foo { barbaz: 10 };
+        thing
+             .
+        thing.
+    }";
+    let path = tmpname();
+    write_file(&path, src);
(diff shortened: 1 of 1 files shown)


johan-sun:
It is so common we put `.` next line in function call chain. (like call iterator adaptors)

PR #472, opened 2016-01-14
Cleanup some warnings from Clippy

mcarton:
For information, here is the [log before and after](https://gist.github.com/mcarton/3496037f45d315bedc9a) with Clippy.
Remaining warnings from Clippy are mostly false positive.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -88,7 +88,7 @@ fn match_fn(m: Match, session: &core::Session, interface: Interface) {
 fn complete(cfg: Config, print_type: CompletePrinter) {
     if cfg.fqn.is_some() {
         return external_complete(cfg);
-    } 
+    }
     complete_by_line_coords(cfg, print_type);
 }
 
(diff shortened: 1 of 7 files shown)


mcarton:
By the way:
[“*Should I return a boxed trait object to make this signature nicer?*”](https://github.com/phildawes/racer/blob/2f9007a623fac7bfb86dc9e729e2ffea2f8206a9/src/racer/matchers.rs#L16)
Clippy [says you should](https://gist.github.com/mcarton/3496037f45d315bedc9a#file-after-L33) do something about that :smile:
```rust
src/racer/matchers.rs:20:55: 20:135 warning: very complex type used. Consider factoring parts into `type` definitions, #[warn(type_complexity)] on by default
src/racer/matchers.rs:20                    local: bool, session: &Session) -> iter::Chain<MChain<MChain<MChain<MChain<MChain<MIter>>>>>, vec::IntoIter<Match>> {
```

phildawes:
Hi @mcarton. These changes look good to me, thanks v much.
Re the chained iterator signature: rust currently doesn't have a good solution to this problem. I've left it the way it is hoping that the 'impl trait' proposal will fix this, otherwise long term we probably want to do something different to chaining iterators.

PR #476, opened 2016-01-16
Use defaults when lib.path or lib.name are not set when searching for main library crate

### LINKED ISSUE #471: The lib crate is invisible from other crates from the same package.
Let's say that we have package `foo` with the `src/lib.rs` containing:
```rust
pub struct Foo { x: i32; }
```

Now, we try to use autocompletion in `src/main.rs`:
```rust
extern crate foo;
use foo::F
```

It should complete to struct `Foo`, but it doesn't work – Racer behaves as it had no idea of the existence of the `foo` crate. The similar issue exists when editing `tests/lib.rs` and probably `examples/*.rs` too.

krdln:
This handles situations when there's no `path` or `name` in `[lib]` section or even when there's no `[lib]` section. Fixes #471

Note that this applies only to situation when searching from within the same package, not general searching of crate root.

### DIFF
--- a/src/racer/cargo.rs
+++ b/src/racer/cargo.rs
@@ -240,15 +240,34 @@ fn find_src_via_tomlfile(kratename: &str, cargofile: &Path) -> Option<PathBuf> {
 
 
     // is it this lib?  (e.g. you're searching from tests to find the main library crate)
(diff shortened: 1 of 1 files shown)


phildawes:
Oh nice, thanks @krdln

PR #477, opened 2016-01-16
Add URLs to Cargo.toml

killercup:
This way, the repository (which is also the homepage) will be linked on crates.io.


phildawes:
Cool, thanks @killercup

Issue #354, opened 2015-08-24
Stack overflow on completion (when function and mod are named the same)


bluss:
I have a project using itertools. Completing for example `itertools::I` works as expected, but `itertools::` completes to nothing. Further research shows it stack overflows when given this completion. [RUST_LOG=debug output is here](https://gist.github.com/bluss/2973167c5cd022dcf6ca).

It was me that replaced $HOME into the logfile.

bluss:
A closer look at the log reveals that it looks like the completion for `itertools::times` that runs in an infinite loop. It's just a function. There's a type of the same name but titlecase (`Times`).

bluss:
Oh and itertools::times is both a private module and a function... that must be it

Here's a minimal testcase:

`testcase.rs`
```rust
pub use times::times;

mod times {
    pub fn times() -> Times { Times }
    pub struct Times;
}

fn main() {
    self::
}
```

stack overflow with completion command to complete at `self::`.
```
racer complete 9 10 testcase.rs
```

phildawes:
Thanks very much for the concise testcase @bluss, I'll take a look

phildawes:
I took a quick look, it looks like even
```
pub use times::times;
```
with `racer find-definition 1 17 testcase.rs` is enough to trip this issue. The problem is that racer searches for `times::times`, decomposes the search into 'find(find("times"), "times")', and then immediately finds the statement again when searching for the outer 'times' module, which triggers a search for `times::times`....

I'm going to have to think hard about the best way to solve this issue.

bluss:
Triage: Still exists in racer as of today, after 1.2.10 at commit b36a9a6af862a63db9397e76abc3cb2013751b8d

itertools suffers from this in version 0.5.0 and [worked around](https://github.com/bluss/rust-itertools/commit/670e844272780d9d9bc763b061e73d5813777484) this issue in release 0.5.1

jwilm:
Hi @bluss,

I've [spent a few hours](https://github.com/phildawes/racer/pull/479) looking into this issue in the past, but it's not particularly straight-forward to fix. Basically, the nameres just recurses indefinitely, and some kind of recursion guard needs to be implemented. Hopefully I'll have time to dig into it again soon; I was hoping for last weekend but that didn't pan out.

This is the number one issue in Racer right now. Nothing else breaks in such a spectacular fashion.

bluss:
[...]

PR #479, opened 2016-01-16
[NOT READY] Stack overflow bug

jwilm:
This is just a repro right now. Will be working on a fix later today. More info here https://github.com/jwilm/racerd/issues/16.

[Review on Reviewable](https://reviewable.io/reviews/phildawes/racer/479)

### DIFF
--- a/tests/system.rs
+++ b/tests/system.rs
@@ -248,6 +248,42 @@ fn follows_use() {
 }
 
 #[test]
+fn completes_out_of_order_mod_use_with_same_fn_name_as_mod() {
+    let cmd_mod="
+    pub fn cmd() {}
+    ";
+
+    let stuff_mod="
+    pub use cmd::cmd;
+
+    mod cmd;
+    ";
+
+    let lib="
+    mod stuff;
+
+    use stuff::
+    ";
+
+    let path = Path::new("lib.rs");
+    write_file(&path, lib);
+    fs::create_dir("stuff").unwrap();
+    write_file(&Path::new("stuff/mod.rs"), stuff_mod);
+    write_file(&Path::new("stuff/cmd.rs"), cmd_mod);
+
+    let pos = scopes::coords_to_point(lib, 4, 15);
+    let cache = core::FileCache::new();
+
+    let got = complete_from_file(lib, &path, pos, &core::Session::from_path(&cache, &path, &path)).nth(0).unwrap();
+    assert_eq!(got.matchstr,"cmd".to_string());
+
+    fs::remove_file(&path).unwrap();
+    fs::remove_file(&Path::new("stuff/mod.rs")).unwrap();
+    fs::remove_file(&Path::new("stuff/cmd.rs")).unwrap();
+    fs::remove_dir("stuff").unwrap();
+}
+
+#[test]
 fn follows_use_as() {
     let src2="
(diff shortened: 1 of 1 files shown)


jwilm:
I've looked at this on and off for the last few days, and am not even close to a solution. However, I've spent a bunch of time reading racer code and have a few observations that I would like to write down as a sort of rubber duck debugging for myself.

The code I'm running through racer is the following.

```rust
use batman::batman;

mod batman {
    pub fn batman() {}
}

fn main() {
    self::
}
```

The completion is running at (8, 10), end of the `self::` line. Racer starts off searching the function body for any completions. Upon not seeing any, the scope expands to the whole file. The first thing encountered is `use batman::batman;` Eventually, the `match_use` function recognizes the line and calls `resolve_path` with `batman::batman` as the path. This is where things really start to go wrong. `resolve_path` sets the parent_path to `batman`, recurses, sees that there is no parent path, calls `resolve_name`, runs the matchers, and hits `match_use` again.

Expected behavior at this point should be for resolve_path to ignore this use line and find the `batman` module.

jwilm:
@phildawes I think you've thought about this problem a little bit (judging from #354). Would you mind sharing? I tried implementing a pretty hacky workaround to prevent the recursion starting in `match_use`, but that didn't entirely fix the problem, and it was absolutely gross :stuck_out_tongue_closed_eyes:.

jwilm:
Actually, it didn't work because of a programming error on my part. I can push the hack later since it might aid discussion (needs a fix, and I need to run).

jwilm:
What about passing a `SearchSet` into `resolve_path`, `resolve_name`, and anything that calls them? Checks for this could be O(1), and it would be pretty general. Anything that has issues with recursion could check the current `Search` membership of the set.

phildawes:
Hi @jwilm. Sorry for not getting back to you. Yes I think this is a good idea. A while back racer had a TLS stack of searches for doing preventing infinite recursion, but then the problem it was solving got solved a different way and it wasn't needed any more.
I think I prefer passing a searchset to TLS because it's more obvious what's going on.

jwilm:
@phildawes thanks for the reply! Why TLS? Is this just to avoid additional arguments to the variety of matcher/search/resolve functions? I had also considered stuffing the search set onto the Session object since that's pretty much always available.

phildawes:
It was originally TLS because it was a general recursion protector - there were a few ways racer could end up recursing on itself

allan-simon:
any  news on this PR or how one can help ? if I understand correctly this PR would help with racerd stack-overflowing when doing completion with redis crate, which I'm really interested in as all my rust project these days are related with redis :)

jwilm:
@allan-simon I feel your pain. We use the redis crate at work. Having not had time to finish this PR, I've resorted to just not using racer with the redis crate.

To finish off this PR, some sort of recursion protection needs to be implemented. Using either a thread-local or stack based set to track search parameters would help in detecting it. After detection, the search function would need to return whatever results had been collected so that the rest of the algorithm can advance.

Hmm, looks like there's no changes on this PR. I've got some work on my home computer, but it's mostly throw away at this point from diagnosing the problem.

Sorry I've sat on this. Haven't had the time/motivation to finish it.

PR #484, opened 2016-01-21
fix when there's no dependencies in cargo file

jaxx:
Fixes #469 panic when there's no dependencies in ```cargo.toml``` file. Fixes also code completion when importing standard library.


phildawes:
Merged, thanks @jaxx

PR #488, opened 2016-01-24
Add RAII helpers TmpDir and TmpFile for tests

jwilm:
Tests in system.rs previously managed creating a temporary name, writing a file, and removing it. The steps associated with creating and removing these files and directories is now abstracted from test writers. Additionally, since these types have Drop impls for cleanup, the files and directories they manage will be removed even on test failure.

The rand crate was added to add some entropy to test names to reduce the probability of name collisions.


phildawes:
This looks great, thanks @jwilm

PR #498, opened 2016-02-02
Replace Arena/Ref caching by a simple hash map storing Rc values.

birkenfeld:
This is an alternate for #497.

The cache handling is much simpler, with the downside that it can't hand out references, it hands out `Rc` clones. That means that the callers have to convert them to `Src` all over the place, which is the reason for the added `as_src()`.  (This used to be `as_ref()` but that is already a `Rc` method.)

I can't see a difference in speed; the additional refcount operations are likely completely insignificant.

A caveat is that this cache is not threadsafe, which might matter if a daemon process should handle concurrent requests (not sure that makes sense though). Switching to `Arc` and `Mutex` might be enough to make it `Sync`.

/cc @jwilm

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -120,7 +120,7 @@ enum CompletePrinter {
 }
 
 #[cfg(not(test))]
-fn cache_file_contents_from_stdin<'a>(file: &PathBuf, cache: &'a core::FileCache<'a>) {
+fn cache_file_contents_from_stdin(file: &PathBuf, cache: &core::FileCache) {
     let stdin = io::stdin();
 
     let mut rawbytes = Vec::new();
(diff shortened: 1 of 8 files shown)


jwilm:
Wow, you've been busy! I like the simplicity of the Rc implementation, and it has a clear path for becoming thread safe if the need arises. FWIW, when I first started working on the Racer code, I had a lot of difficulty reasoning about the caching and the references. This version would eliminate that for future contributors.

The one downside I see to Rc is that, theoretically, a stale file could be used as part of an operation.

jwilm:
I'm thinking the Rc version of this is the way to go. So simple :).

phildawes:
Still not completely sure which alternative is best, but merging this because it was the simplest.
Thanks so much for this @birkenfeld!

PR #500, opened 2016-02-04
Show type information for field completions

Manishearth:
fixes https://github.com/jwilm/racerd/issues/26

Might want to accompany this with a version bump and publish since racerd wants to use it.

r? @phildawes

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -132,6 +132,35 @@ pub enum Ty {
     TyUnsupported
 }
 
+impl fmt::Display for Ty {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            Ty::TyMatch(ref m) => {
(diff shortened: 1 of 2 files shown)


Manishearth:
Screenshot of it working:

screenshot from 2016-02-05 04 42 10

Note that non scalar primitive types do not show up properly. On completing the following code:

```rust
struct Foo<'a> {
 reference: &'a u8,
 array: [u8; 5],
 slice: &'a [u8],
}

fn foo(x: Foo) {
 x.
}
```

the following completions show up. We might want to add a TyPrimitive to handle `&T`, `&[T; N]` and `&[T]` (not sure if we should attempt to work with function types).

screenshot from 2016-02-05 04 45 19

phildawes:
Awesome work, thanks @Manishearth

jwilm:
Huzzah! :100:

phildawes:
(Will do the racer release tonight when I get home)

PR #502, opened 2016-02-05
Fix formatting of contextstr for refs/vecs

jwilm:
Given the following example,

```rust
struct Foo<'a> {
    reference: &'a u8,
    array: [u8; 5],
    slice: &'a [u8],
}

fn foo(x: Foo) {
    x.
}
```

The contextstr for completions should be `&u8`, `[u8; 5]`, and `&[u8]` for each `Foo` member,  espectively. This patch fixes behavior where contextstr was `u8`, `array`, and `slice` instead.

Variants were added to `core::Ty` to represent ref ptrs, fixed length vecs, and vecs. The `Display` impl for `Ty` was updated to format them as expected.

Perhaps questionably, these new variants hold a `Box` for another `Ty`. This makes awkward anywhere that needs to destructure the variant and pass the nested type as an owned object.

/cc @Manishearth

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -129,6 +129,9 @@ pub enum Ty {
     TyMatch(Match),
     TyPathSearch(Path, Scope),   // A path + the scope to be able to resolve it
     TyTuple(Vec<Ty>),
+    TyFixedLengthVec(Box<Ty>, String), // ty, length expr as string
+    TyRefPtr(Box<Ty>),
+    TyVec(Box<Ty>),
     TyUnsupported
 }
 
@@ -153,6 +156,21 @@ impl fmt::Display for Ty {
                     }
                 }
                 write!(f, ")")
+            }
+            Ty::TyFixedLengthVec(ref ty, ref expr) => {
+                try!(write!(f, "["));
+                try!(write!(f, "{}", ty));
+                try!(write!(f, "; "));
+                try!(write!(f, "{}", expr));
+                write!(f, "]")
+            }
+            Ty::TyVec(ref ty) => {
+                try!(write!(f, "["));
+                try!(write!(f, "{}", ty));
+                write!(f, "]")
+            }
+            Ty::TyRefPtr(ref ty) => {
+                write!(f, "&{}", ty)
             }
             Ty::TyUnsupported => {
                 write!(f, "_")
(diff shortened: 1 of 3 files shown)


src/racer/core.rs
```diff
@@ -648,17 +666,29 @@ pub fn complete_from_file(src: &str, filepath: &path::Path,
...
-                }
+                complete_field_for_ty(ty, searchstr, SearchType::StartsWith, session, &mut out);
             });
         }
     }
     out.into_iter()
 }
 
+fn complete_field_for_ty(ty: Ty, searchstr: &str, stype: SearchType, session: &Session, out: &mut Vec<Match>) {
+    // TODO would be nice if this and other methods could operate on a ref instead of requiring
```
Manishearth: We may at some point want to handle `String` and `Vec` here (perhaps a special case?), too, since they get a lot of methods out of autoderef through slices.

Manishearth:
Note that you can destructure `box` on nightly (`Pattern(box inner) => ...`).  But then racer won't work on stable.

You should also fix up the `Debug` impl.

Aside from that. looks good to me.

jwilm:
Does destructuring box on nightly skip the copy? Does it consume the box? Basically, where can I read more about this? :) thanks for sharing

Will fix the debug impl tonight.

Manishearth:
It just dereferences the box. Move by default, but you can combine it with `ref` to make it take a reference. Basically, `box` means "match this pattern as `x`, and gimme `*x`" (except right now it only works for boxes and not generic derefable things, but that may change), and `ref` means "match this pattern as `x`, and gimme `&x`". `ref box` gets you `&*x`.


`box` is a rather old piece of syntax that is a holdover from the `~` days. It can be used to create boxes (`let x = box 5` -- this is slightly different from `Box::new` in that the `5` is written directly to the memory slot -- this matters for large arrays being boxed) and also to pattern match them. It wasn't removed even after `~` became `Box`, and now it's going to perhaps be used in generic placement new. I don't know if there are any concrete plans of what we'll do with `box` patterns itself, but I suspect it will eventually be able to deref arbitrary patterns (there's a slight issue here, in that `*x` on `Box<T>` is not the same as `*x` on `SomeOtherDereferenceable<T>`. The former always moves unless paired with a `&`, the latter always borrows.)

jwilm:
Thanks for the explanation. It seems like a move in this case still requires a copy of the underlying memory since the new owner would be stack allocated. Is that right?

Manishearth:
Yeah, but the `Ty` struct isn't large.

jwilm:
`Ty` isn't large, but whatever I'm working on next might be :grin:

Manishearth:
Ah. Note that you can always use `&*x` (or just `&x`, since deref coercions should make it happen) to obtain/use an `&T`, which will be able to call `&self` methods on `T` just fine. It's only going to copy when you try to move.

jwilm:
@Manishearth `Ty` derives `Debug` - what do you mean fix it up?

Manishearth:
my bad, that's Path I'm thinking of

phildawes:
Thanks @jwilm

phildawes:
released 1.2.4 to include this

PR #516, opened 2016-03-19
Extern crate match fix

BGluth:
Racer was not picking up the names of actual crates from `extern crate {}` lines.

So for example this code would not match the `ecs` crate:
***************************
`extern crate ecs`
`ec` _(run racer here)_
***************************
This was due to the external crate matching code placing an unneeded `;` at the end of the searchstr it was trying to match.

`if txt_matches(search_type, &format!("extern crate {};", searchstr), blob)`

So it would try to find "extern crate ec;" inside of "extern crate ecs;" and not find anything.


phildawes:
Thanks v much @BGluth, merged!

PR #518, opened 2016-03-22
use standard library path.exists() and path.is_dir() functions

jaxx:
Replace util::path_exists() with path.exists() and util::is_dir() with path.is_dir() functions as those are now stabilized :)


phildawes:
Thanks @jaxx, merged

PR #521, opened 2016-03-28
Multiple use glob fix

BGluth:
If a file contained multiple use globs in the available scope, then only the use glob processed first would be evaluated.

e.g.
****************
extern crate A;
extern crate B;

use A::*;
use B::*;

text <-- (Run Racer Here)
****************

- Running Racer here would only give matches for crate A.
- This was due to match_use not setting "ALREADY_GLOBBING" back to "None" once it finished working on the glob before exiting on the early return statement, which would prevent any other use globs from being processed afterwards.


phildawes:
This is awesome, thanks @BGluth

PR #526, opened 2016-04-05
Fix #470

### LINKED ISSUE #470: Improve README.md
Two remarks on the front page README:

- If I am not mistaken, points 3, 4 and 5 of the **From sources** section should be actually outside of the section, ie. common to this one and to the cargo install section. In this case, the command should have no path. Alternatively, duplicate the points and adapt to the other section.
- The [Atom package](https://atom.io/packages/racer) should be mentioned with the other plugins. Perhaps there are others, too.

bwinterton:
I have added a configuration section to the README.md which includes the
information that is common to all installation methods (downloading the
rust src). I have also added a step to add the binary to your `PATH`
from the source installation method. I have added a section about the
racer atom package as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,15 +24,19 @@ As mentioned in the command output, don't forget to add the installation directo
 
 2. ```cd racer; cargo build --release```.  The binary will now be in ```./target/release/racer```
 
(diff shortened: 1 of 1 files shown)


phildawes:
Thanks @bwinterton

PR #527, opened 2016-04-06
Add note for rustup/multirust-rs users

Bobo1239:
Also see https://github.com/rust-lang-nursery/multirust-rs/issues/46

The problem is that both [rustup/multirust-rs](https://github.com/rust-lang-nursery/multirust-rs) and [multirust](https://github.com/brson/multirust) use the .multirust directory but only multirust has separate cargo homes for different toolchains.

After https://github.com/rust-lang-nursery/multirust-rs/issues/247 is resolved, racer can (more easily) differentiate between the two cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,12 @@ As mentioned in the command output, don't forget to add the installation directo
 
    ```./target/release/racer complete std::io::B ```  (should show some completions)
 
+### Note for [rustup/multirust-rs](https://github.com/rust-lang-nursery/multirust-rs) users
+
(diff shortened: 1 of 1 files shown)


brson:
The multirust metadata situation is a mess. Sorry. You can distinguish the multirust.sh and multirust-rs cases today by the contents of the `version` file, which is '2' for multirust and '12' for multirust-rs.

phildawes:
Cool, thanks @Bobo1239

PR #530, opened 2016-04-14
Fix issue resolving local path dependencies

### LINKED ISSUE #529: Local path lookup semi-broken
I ran into an issue where having a `dependencies.whatever.path` in my Cargo.toml would result in no completions/definitions, but putting the same path in a `.cargo/config` file fixed the issue.

This needs a minimal test case; will try and come back later and provide one.

jwilm:
Local path dependencies sometimes (usually) failed to resolve. There
turned out to be an issue when processing the cargo file where if a
dependency didn't have a `path` attribute, the get_local_packages
function returned None. Now, the function only returns None when there
is no parent directory for the cargo file.

Resolves #529.

@phildawes would be nice to release a cargo update for this since it fixes a pretty significant bug.

@Manishearth this might fix some issues you've seen completing within servo

---
This change is [Reviewable](https://reviewable.io/reviews/phildawes/racer/530)

### DIFF
--- a/src/racer/cargo.rs
+++ b/src/racer/cargo.rs
@@ -336,6 +336,7 @@ fn find_src_via_tomlfile(kratename: &str, cargofile: &Path) -> Option<PathBuf> {
 
     // if no dependencies are found
     if local_packages.is_empty() && local_packages_dev.is_empty() {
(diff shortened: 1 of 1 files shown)


phildawes:
Thanks @jwilm. Will do a release tonight

PR #531, opened 2016-04-14
Fix cargo lock lookup

jwilm:
The main loop of `racer::cargo::get_cargo_packages` previously used `otry!` extensively which would cause the entire function to return None in the case that any of the crates in the package list didn't have all of the desired properties. `otry!` usage in this function was replaced with an `unwrap_or_continue!` macro which yields the value of some or `continue`s which skips the current package.

The remaining changes are cleanup and logging additions. The main commit to look at for review purposes is 27602934fe6d52525016c18e5b1abfa07fb2aec3. There was also some minor refactoring in 6eaefb2ad83b47312e4c089963ef5c7d6a55536d.

---
This change is [Reviewable](https://reviewable.io/reviews/phildawes/racer/531)

### DIFF
--- a/src/racer/cargo.rs
+++ b/src/racer/cargo.rs
@@ -259,7 +259,7 @@ fn get_versioned_cratefile(kratename: &str, version: &str, cargofile: &Path) -> 
                 if let Some(path) = reader
                     .map(|entry| entry.unwrap().path())
                     .find(|path| path.to_str().unwrap().starts_with(start_name)) {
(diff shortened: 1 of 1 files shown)


jwilm:
Relevant - I think this was only a problem when there was a local path dependency. After applying these changes, the added `debug!("get_cargo_packages skipping {}", package_name);` only printed such a dependency.

phildawes:
Thanks @jwilm, merged

PR #534, opened 2016-04-25
Replace hyphens with underscores when constructing a crate name from a package name

thequux:
This follows cargo's behaviour, making code completion work on packages like `lmdb-rs`.


thequux:
I just rebased onto the latest head to fix the merge conflict.

phildawes:
Thanks @thequux, this is great.

Issue #415, opened 2015-10-26
Would it be possible for racer to integrate with rustdoc?


7sharp9:

bruno-medeiros:
+1 , would be a great feature.

phildawes:
Agreed. Currently all my spare time is going into figuring out how to use rustc to get perfect completions so I probably won't get to this for a while I'm afraid.

estebank:
I have a small hacked together version of this, I'll be sending a pull request later today.

I noticed that for some reason in HEAD `complete-with-snippets` is returning the same thing as `complete`. I'll try to get it working with vim-racer as well to present a full feature.

I modified `core::Match` to add a `docs: String` field. Let me know if you disagree with doing that for some reason, @phildawes.

PR #535, opened 2016-05-02
Add function and struct docs to `Match` type

estebank:
Add function and struct docstring to `Match` type so that `complete-with-snippets` can include the docstring. Prospective fix to #415.

Make it so that `complete-with-snippets` works when used for a cli provided complete string.


phildawes:
Thanks @estebank. I ran this today without issue, merged!

PR #538, opened 2016-05-04
Update to syntex 0.32

jwilm:
Apparently Racer doesn't use any APIs that changed, so this is purely a
version bump. Tests passed on my machine.

---
This change is [Reviewable](https://reviewable.io/reviews/phildawes/racer/538)


phildawes:
Thanks @jwilm

PR #539, opened 2016-05-05
clippy recommendations

jaxx:


src/racer/ast.rs
```diff
@@ -437,10 +434,7 @@ impl<'c, 's, 'v> visit::Visitor<'v> for ExprTypeVisitor<'c, 's> {
         debug!("visit_expr {:?}", expr);
         //walk_expr(self, ex, e)
         match expr.node {
-            ExprKind::Unary(_, ref expr) => {
-                self.visit_expr(expr);
-            }
-            ExprKind::AddrOf(_, ref expr) => {
+            ExprKind::Unary(_, ref expr) | ExprKind::AddrOf(_, ref expr) => {
```
jwilm: Can this remain as one variant per line? Eg

```rust
ExprKind::Unary(_, ref expr) |
ExprKind::AddrOf(_, ref expr) => {
    // ...
}
```

I find this version to be more readable.

src/racer/typeinf.rs
```diff
@@ -228,10 +227,7 @@ pub fn get_type_of_match(m: Match, msrc: Src, session: &Session) -> Option<core:
         core::MatchType::For => get_type_of_for_expr(&m, msrc, session),
         core::MatchType::FnArg => get_type_of_fnarg(&m, msrc, session),
         core::MatchType::MatchArm => get_type_from_match_arm(&m, msrc, session),
-        core::MatchType::Struct => Some(core::Ty::TyMatch(m)),
-        core::MatchType::Enum => Some(core::Ty::TyMatch(m)),
-        core::MatchType::Function => Some(core::Ty::TyMatch(m)),
-        core::MatchType::Module => Some(core::Ty::TyMatch(m)),
+        core::MatchType::Struct | core::MatchType::Enum | core::MatchType::Function | core::MatchType::Module => Some(core::Ty::Match(m)),
```
jwilm: Ditto variants per line

jwilm:
This LGTM. For anyone else curious about the single character replacements, the explanation of the clippy lint:

> **Default level:** Warn
> **What it does:** This lint checks for string methods that receive a single-character str as an argument, e.g. _.split("x").
> **Why is this bad?** Performing these methods using a char is faster than using a str.
> **Known problems:** Does not catch multi-byte unicode characters.

jaxx:
Thanks @jwilm! Replaced match arms as you suggested.

jwilm:
Thanks for fixing all the clippy lints!

phildawes:
The single unicode char split being faster than a str split is surprising to me. I need to take a look at the source to see why!

phildawes:
Thanks @jaxx + @jwilm

PR #540, opened 2016-05-08
Escape `;` to split the MATCHes properly

estebank:
Both the signature and the documentation can contain the separator
character (either `;` or `\t`). Escape it so that client applications
can split the MATCHes properly.

Remove first chars from docs without losing indent.


phildawes:
Thanks @estebank

PR #542, opened 2016-05-09
Add module level docs and unittests

estebank:
[How it currently looks on (a pull request on) vim-racer](https://github.com/racer-rust/vim-racer/pull/42):

### DIFF
--- a/src/racer/matchers.rs
+++ b/src/racer/matchers.rs
@@ -21,7 +21,7 @@ pub fn match_types(src: Src, blobstart: usize, blobend: usize,
                    search_type: SearchType,
                    local: bool, session: &Session) -> iter::Chain<MChain<MChain<MChain<MChain<MChain<MIter>>>>>, vec::IntoIter<Match>> {
     let it = match_extern_crate(&src, blobstart, blobend, searchstr, filepath, search_type, session).into_iter();
-    let it = it.chain(match_mod(src, blobstart, blobend, searchstr, filepath, search_type, local).into_iter());
+    let it = it.chain(match_mod(src, blobstart, blobend, searchstr, filepath, search_type, local, session).into_iter());
     let it = it.chain(match_struct(&src, blobstart, blobend, searchstr, filepath, search_type, local).into_iter());
     let it = it.chain(match_type(&src, blobstart, blobend, searchstr, filepath, search_type, local).into_iter());
     let it = it.chain(match_trait(&src, blobstart, blobend, searchstr, filepath, search_type, local).into_iter());
@@ -279,7 +279,7 @@ pub fn match_extern_crate(msrc: &str, blobstart: usize, blobend: usize,
 
 pub fn match_mod(msrc: Src, blobstart: usize, blobend: usize,
                  searchstr: &str, filepath: &Path, search_type: SearchType,
-                 local: bool) -> Option<Match> {
+                 local: bool, session: &Session) -> Option<Match> {
     let blob = &msrc[blobstart..blobend];
     if let Some(start) = find_keyword(blob, "mod", searchstr, search_type, local) {
         debug!("found a module: |{}|", blob);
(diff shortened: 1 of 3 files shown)


src/racer/matchers.rs
```diff
@@ -298,7 +300,7 @@ pub fn match_mod(msrc: Src, blobstart: usize, blobend: usize,
                 contextstr: filepath.to_str().unwrap().to_owned(),
                 generic_args: Vec::new(),
                 generic_types: Vec::new(),
-                docs: String::new(),  //TODO: fix
+                docs: String::new(),
```
jwilm: I realize this was added in a different PR, but this should probably hold an `Option<String>` and initialize to `None` here.

src/racer/matchers.rs
```diff
@@ -311,6 +313,20 @@ pub fn match_mod(msrc: Src, blobstart: usize, blobend: usize,
                 searchdir.push(&s);
             }
             if let Some(modpath) = get_module_file(l, &searchdir) {
+                let f = File::open(&modpath);
+                let mut file = f.unwrap();
+                let mut rawbytes = Vec::new();
+                file.read_to_end(&mut rawbytes).unwrap();
```
jwilm: This should use the `Session` object to read files.

src/racer/matchers.rs
```diff
@@ -658,11 +674,38 @@ fn find_doc(msrc: &str, blobend: usize) -> String {
...
+        })
+        .collect::<Vec<_>>()
+        .join("\n")
+}
+
+fn find_mod_doc(msrc: &str, blobstart: usize) -> String {
+    let blob = &msrc[blobstart..];
+
+    blob.lines()
+        .take_while(|line| {
```
jwilm: Each of the blocks here calls `line.trim()`. Would it be better to have an initial `map(|line| line.trim())` to save some cycles in subsequent steps?

src/racer/matchers.rs
```diff
@@ -658,11 +674,38 @@ fn find_doc(msrc: &str, blobend: usize) -> String {
...
+        .map(|line| {
+            let l = line.trim();
+            if l.len() >= 4 {  // Remove "/// "
+                String::from(l[4..].to_owned())
+            } else {
+                String::new()
+            }
+        })
         .collect::<Vec<_>>()
         .join("\n")
```
jwilm: There's several allocations happening in here including 2 from `String::from(l[4..].to_owned())` (the `String::from` and `to_owned` are redundant), and the `collect::<Vec<_>>()`. It may be more efficient to save this iterator to a temporary and then build a string like

```rust
let mut doc = String::new();
for line in doc_lines {
    doc.push_str(line);
    doc.push_str("\n");
}
```

jwilm:
Thanks for putting this together! Getting module docs will be nice, and having tests for this stuff is crucial to keep it working. Really appreciate the work here.

src/racer/matchers.rs
estebank: Fixed.

src/racer/matchers.rs
estebank: Done.

src/racer/matchers.rs
estebank: Done for `find_mod_doc`, not for `find_doc`.

(The easy fix in `find_doc` of assigning the iterator to a variable ends up in a `"variable does not live long enough"` error, and you need at least one arguably unnecessary `collect` to be able to `rev`erse the iterator.)

src/racer/matchers.rs
estebank: Isn't an empty string arguably the same as a None? There's never going to be a zero length doc string, so assuming that an empty string means no documentation sounds like a fair trade off between correctness and ease of use of the interface. Would still be happy to hear a different argument :)

jwilm: I was thinking that `String::new` allocated, but that's simply not true. Thanks for making me look it up! The optionless version is a much nicer API.

phildawes:
Awesome, thanks @estebank

PR #543, opened 2016-05-11
reformat contextstr of multiline signatures

### LINKED ISSUE #512: racer only returns the first line of function signatures
Given the following code:

```rust
fn add_nums(x: i64,
            y: i64)
            -> i64 {
    x + y
}

fn main() {
    add_nums(1, 2);
}
```

Racer returns:

```
$ racer complete 8 7 foo.rs     
PREFIX 4,7,add
MATCH add_nums,1,3,foo.rs,Function,fn add_nums(x: i64,
END
```

Ideally racer would return the whole function signature.

estebank:
Fixes #512.

```
fn add_nums(x: i64,
            y: i64)
            -> i64 {
    x + y
}

fn main() {
    add_nums(1, 2);
}
```
Racer now returns:
```
racer complete 8 7 src/x.rs
PREFIX 4,7,add
MATCH add_nums,1,3,src/x.rs,Function,fn add_nums(x: i64, y: i64) -> i64
END
```
instead of:
```
racer complete 8 7 src/x.rs                                        {master
PREFIX 4,7,add
MATCH add_nums,1,3,src/x.rs,Function,fn add_nums(x: i64,
END
```

### DIFF
--- a/src/racer/matchers.rs
+++ b/src/racer/matchers.rs
@@ -221,6 +221,17 @@ pub fn first_line(blob: &str) -> String {
     (&blob[..blob.find('\n').unwrap_or(blob.len())]).to_owned()
 }
 
+/// Get the match's cleaned up context string
+///
+/// Strip all whitespace, including newlines in order to have a single line
+/// context string.
+pub fn get_context(blob: &str, context_end: &str) -> String {
+    (&blob[..blob.find(context_end).unwrap_or(blob.len())])
+        .split_whitespace()
+        .collect::<Vec<_>>()
+        .join(" ")
+}
+
 pub fn match_extern_crate(msrc: &str, blobstart: usize, blobend: usize,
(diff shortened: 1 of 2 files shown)


src/racer/matchers.rs
```diff
@@ -221,6 +221,17 @@ pub fn first_line(blob: &str) -> String {
     (&blob[..blob.find('\n').unwrap_or(blob.len())]).to_owned()
 }
 
+/// Get the match's cleaned up context string
+///
+/// Strip all whitespace, including newlines in order to have a single line
+/// context string.
+pub fn get_context(blob: &str, context_end: &str) -> String {
+    (&blob[..blob.find(context_end).unwrap_or(blob.len())])
```
jwilm: I think clippy will prefer `unwrap_or_else(|| blob.len())`

jwilm:
Looks good!

phildawes:
Great, thanks @estebank

Issue #496, opened 2016-02-01
racer panic on string


letheed:
The following code:

``` rust
fn main {
    let f = File::open("tests/m256..257.txt").unwr
}
```

produces the following panic:

``` shell
$ racer complete 2 50 main.rs
PREFIX 46,50,unwr
bogofile:1:8: 1:10 error: unterminated double quote string
bogofile:1 257.txt")
                  ^~
thread 'searcher' panicked at 'Box<Any>', /home/damien/.cargo/registry/src/github.com-88ac128001ac3a9a/syntex_syntax-0.24.0/src/parse/lexer/mod.rs:1117
stack backtrace:
   1:     0x55e27d260aa0 - sys::backtrace::tracing::imp::write::hb120982054a416e35nu
   2:     0x55e27d2630ab - panicking::default_handler::_$u7b$$u7b$closure$u7d$$u7d$::closure.42840
   3:     0x55e27d262d16 - panicking::default_handler::h7ae2e4523ef4c187IFy
   4:     0x55e27d25298c - sys_common::unwind::begin_unwind_inner::h5fb19481d14902dbDgt
   5:     0x55e27d1265ba - sys_common::unwind::begin_unwind::h11772222310432410183
                        at ../src/libstd/sys/common/unwind/mod.rs:236
   6:     0x55e27d1c4a7c - parse::lexer::StringReader<'a>::next_token_inner::hdcdff1aaac931df1Y3Q
                        at /home/damien/src/lang/rust/racer/<std macros>:3
   7:     0x55e27d1be2bb - parse::lexer::StringReader<'a>::advance_token::ha6797d50ae87f5a9DiQ
                        at /home/damien/.cargo/registry/src/github.com-88ac128001ac3a9a/syntex_syntax-0.24.0/src/parse/lexer/mod.rs:240
   8:     0x55e27d15aeeb - parse::lexer::StringReader<'a>.Reader::next_token::h88596fc7d03833f3L8P
                        at /home/damien/.cargo/registry/src/github.com-88ac128001ac3a9a/syntex_syntax-0.24.0/src/parse/lexer/mod.rs:89
   9:     0x55e27d064ea3 - parse::lexer::Reader::real_token::h6596309516582769722
                        at /home/damien/.cargo/registry/src/github.com-88ac128001ac3a9a/syntex_syntax-0.24.0/src/parse/lexer/mod.rs:42
[...]

PR #545, opened 2016-05-14
fix for #496

jaxx:
This is possible fix for #496. Please review it. I think there's some parentheses missing. But I'm not sure if it does break something else or not. Tests are green though.


jaxx:
Interesting, all tests are green in my machine.

PR #550, opened 2016-05-23
Update README.md

donaldpipowitch:
I think this is clearer now. The URL actually changed and I think the project is now just called `rustup`, while `multirust` isn't maintained anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ As mentioned in the command output, don't forget to add the installation directo
 
    ```racer complete std::io::B ```  (should show some completions)
 
-### Note for [rustup/multirust-rs](https://github.com/rust-lang-nursery/multirust-rs) users
+### Note for [rustup](https://github.com/rust-lang-nursery/rustup.rs) users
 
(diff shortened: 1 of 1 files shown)


phildawes:
Hi @donaldpipowitch  - Thanks for the patch.

I might make a couple of changes once I've figured out what works and what doesn't. So basically rustup.rs is the new multirust, but racer currently isn't working with the new rustup-multirust directory structure?

donaldpipowitch:
> So basically rustup.rs is the new multirust.

Yes, this is how I understand it. From the multirust README:

> Note: multirust is not actively maintained. Migrate to rustup.rs.

For me (Mac OS X, rustup 0.1.12) Racer actually works _without_ setting `CARGO_HOME`! I only had to set `RUST_SRC_PATH`. So maybe this whole paragraph is _obsolete_...?

davidbarsky:
@donaldpipowitch My understanding—and correct me if I’m wrong—is that Racer works only with the standard library, not with third-party crates.

Issue #96, opened 2014-12-20
Correctly handle multiple-use


porglezomp:
It would be nice if line like `std::io::{` would complete the same as `std::io::`
It would be even nicer if, for instance `use std::rc::{Rc,` would recommend everything that `std::rc::` recommends, except it would omit `Rc` since it's already being used.

TedDriggs:
Technically, the following code is valid:

```rust
use std::fmt::{Formatter, Formatter as Fmtr};
```

I have no idea _why_ someone would do this, but since it's allowed I'm reluctant to have rust hide things that have already been imported in the current `use` statement.

porglezomp:
I was a little surprised to get this notification, but as far as I can tell the main issue here was resolved a while ago thanks to @yodalee, so I'm happy to have this issue closed.

PR #557, opened 2016-06-06
remove leading bracket in search string

### LINKED ISSUE #554: Bracket stop completion in use
I try racer completer with
`use core::ops::{Index, IndexMut};`
I found that bracket will stop completion. So I cannot complete Index and IndexMut automatically.
Any hint where can I fix this problem?

yodalee:
The racer complete following pattern normally:
$ racer complete std::io::B

However in usual case in editor, bracket is used to
include more than one module. The leading bracket will
break the completer. The following pattern will get no
results:
$ racer complete std::io::{B

This patch fix this problem by removing leading bracket
in search string, so std::io::{B will work same as
std::io::B

Close #554

### DIFF
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -1088,7 +1088,11 @@ pub fn do_external_search(path: &[&str], filepath: &Path, pos: usize, search_typ
             match m.mtype {
                 Module => {
                     debug!("found an external module {}", m.matchstr);
-                    let searchstr = path[path.len()-1];
+                    // deal with started with "{", so that "foo::{bar" will be same as "foo::bar"
+                    let searchstr = match path[path.len()-1].chars().next() {
+                        Some('{') => &path[path.len()-1][1..],
+                        _ => path[path.len()-1]
+                    };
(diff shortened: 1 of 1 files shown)


phildawes:
Thanks @yodalee !

jwilm:
Does this actually resolve #554?

The example given in that issue was
```rust
use core::ops::{Index, IndexMut};
```
and this patch sounds like it would just fix completion of Index, but not IndexMut.

knight42:
@jwilm Agree.

Now `racer complete 'core::ops::{Add, In'` gives the same result as `racer complete 'core::ops::{Add'`.

Perhaps racer needs to ignore `{(.*,)?` (regexp) ?

yodalee:
I just fix the index actually, and that is what #554 I open asking for. I will take some time looking into the comma issue.
These are all duplicated issue to #96.

yodalee:
I just do some code trace. The input matching string `std::io::{B,` will become searchstr in `do_external_search` as `std::io::{B` the comma is automatically removed. These is cause be `clap` parsing method.

PR #559, opened 2016-06-07
use comma-separated rightmost part as searchstr

yodalee:
In previous version, multiple use is not handle correctly
The comma separated use cannot get right completion.
First the clap prevent comma being parsed into fqn,
we need to enable it by setting `use_delimiter` false.

The searchstr is now being processed as below:
1. If there is comma, take the rightmost part as search string.
2. If there is leading bracket, ignore it.

This will close issue 96 except that multiple use ignore those module already being use.
For example, complete `std::io::{BufReader, B` will still have `BufReader` in output.
Issue #96 suggests that in above case BufReader should not appear. I think this is only tiny `bug` and not very inconvenient to our daily use.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -368,6 +368,7 @@ fn build_cli<'a, 'b>() -> App<'a, 'b> {
             // We start by making 'fqn' the first positional arg, which will hold this dual value
             // of either an FQN as it says, or secretly a line-number
             .arg(Arg::with_name("fqn")
+                .use_delimiter(false)
                 .help("complete with a fully-qualified-name (e.g. std::io::)"))
             .arg(Arg::with_name("charnum")
                 .help("The char number to search for matches")
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -1049,7 +1049,13 @@ pub fn do_external_search(path: &[&str], filepath: &Path, pos: usize, search_typ
     debug!("do_external_search path {:?} {:?}", path, filepath.to_str());
     let mut out = Vec::new();
     if path.len() == 1 {
-        let searchstr = path[0];
+        let mut searchstr = path[0];
+        if let Some(i) = searchstr.rfind(',') {
+            searchstr = &searchstr[i+1..].trim();
+        }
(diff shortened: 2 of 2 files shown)


jwilm:
This is great; thanks for putting a PR together! Any chance you could add a test for this behavior so it doesn't accidentally become broken in the future?

yodalee:
OK, let me try if I can put some test into it after I wake up next morning XD.

yodalee:
I'm not sure that is this `searchstr`
https://github.com/yodalee/racer/blob/f06fec6051d5fa486e4184e95c6ca52895b74fc2/src/racer/nameres.rs#L1052
should truncate leading bracket too?

jwilm:
@yodalee for the test, I would copy one of the existing completion tests, define some `fn` in a module, and have it try and complete from that module in the parent scope. [This test](https://github.com/phildawes/racer/blob/master/tests/system.rs#L638-L655) seems like a good starting point.

yodalee:
I knew that. But the modification I made is somehow different from existing tests. These tests are testing against function `racer::core::complete_from_file`, which has different function call with the function `do_external_search`.

jwilm:
@yodalee `complete_from_file` is the function that should be tested against here. Racer is more interested in system level tests than unit tests.

phildawes:
Hi @yodalee,

Thanks for working on this. @jwilm is right, `complete_from_file` is the one you want to code against since that is the logic that IDEs will be (indirectly) running.

E.g. In this case, running
`% racer complete std::io::{B` 
...completes ok, but unfortunately actually running racer from an editor still fails. E.g. try creating a file and then running racer against it on the command line:

/tmp/a.rs:
`use std::io::{B`

then `% racer complete 1 15 /tmp/a.rs`

yodalee:
Yeah, that is what I concerned. I make the command `racer complete std::io::` handling comma and bracket normally. However, that is not the command used in common editor. They are different function call actually.
So this PR can be merged but it doesn't make any difference lol

TedDriggs:
This seems to work, and a related PR was merged. @yodalee, please let me know if this being closed is an error.

Issue #561, opened 2016-06-10
racer doesn't handle importing multiple name from the same module


AsafMah:
Basically,
typing `use std::cell::`
gives you what you would excpect, Cell, Refcell, etc.

But typing `use std::cell::{` doesn't give you any completions, even though 
`use std::cell::{Cell, RefCell};` is a legitimate line.

As far as I know it produces the same result in all editors, and directly in the command line.

yodalee:
You may try the newest master branch. The leading bracket is handled now #557. The comma (to complete Refcell)  is waiting some testings #559

yodalee:
I dig into this issue for a while, here is some cause I think.
There are three different issues: leading bracket, comma-separated, and space after comma.
In `core.rs` it will call `get_start_of_search_expr`, this will cut down string represents a expression. Since starting bracket is not part of expression, the `use std::cell::{` will become matching a empty string, and then it matches nothing.
I tried to add leading bracket into part of expression ( by adding it to `utils::is_search_expr_char` ), allowing it become `searchstr`. Then in `nameres.rs::search_scope`, I remove bracket from searchstr if it is started with it. This can solve problem one.
I push a branch to my repository, here is the commit diff:
https://github.com/phildawes/racer/commit/9e0825a7fe78e734c5432acd0d00aacdea49c288

I don't like this solution, though. It cannot handle comma-separated and space after comma. If we tried to add comma and space into `utils::is_search_expr_char`, many tests case will be broken. It's not surprised because space shouldn't be part of expression.
Just my feeling. I think more ultimate solution will lies in somewhere in `ast parser` or `nameres::search_scope_headers`. By viewing the running log, the `use` line will trigger a parser in ast by calling `ast::parse_use`. The parser will parse the line with `parse_crate_mod in syntex_syntax`, though it usually failed and return None. There may have some way to let it parse correctly, then the code in `matchers::match_use` can work normally.

I haven't got the whole view of how racer works, and it's just my thought. Maybe someone who knows racer more can give me some hints about how to fix the issues or at least where should I look at.

cc @phildawes @jwilm

TedDriggs:
Thanks to @yodalee for fixing this!

PR #563, opened 2016-06-18
Handle multiple use correctly in complete from file

yodalee:
The Cause:
In origin design, the expression before complete points will be
splitted out, with space or special character as delimiter.
This, however, is not correct with multiple use.
Here is typical syntax for `use`:
$ use contextstr::{A, B, C, searchstr} or contextstr::{searchstr
$ contextstr::searchstr or ::contextstr::searchstr

In the first case, both bracket, comma and space will stop parsing
expression, so we will get incomplete context string to resolve.
For example, in the following case, we will get empty context since
bracket stop parsing:
$ use std::io::{B

The Solution:
Instead of expression, if complete type is Path. We try to parse
entire line contains the complete points. The lines will then
goes through these processes:
1. trim the leading spaces.
2. split by semicolon then take the rightmost part
   this prevents user write multiple lines in one lines, such as:
   use std::io::BufReader; b.app
3. If the line is started with `use ` or `::` then we should
   consider whole line as completion. Otherwise expression parsing
   result is corrected.

Also, in function resolve_path, the searchstr is the last part splitted
by `::`, it may contains strings like:
`{searchstr` or `{A, B, searchstr`

The searchstr is processed with steps:
1. The searchstr is split by comma, and then take the rightmost part.
2. If the searchstr is started with '{', simply remove it.
By this we can handle multiple use correctly.

TODO:
In the function `core::complete_from_file`, there is room for
improvement. Firstly it cut down expression for searching, determine
its type is `Path` or `Field`. These steps are often unused with
complete `Path` type. We have to re-parsed the line contains the
complete points, creating redundant steps.
The steps in `core::complete_from_file` need some refactor, to clarify
data preparing steps and remove redundant source code parsing.

These will close issue #561 #96
Also: PR #559 can be merged though it deals with same problem with totally different calling path.


yodalee:
Two new cases test broken due to RUST_SRC_PATH didn't set to correct position. The complete on `std::io::B` cannot get correct results.

phildawes:
Hi @yodalee,

Thanks for working on this!

Re tests: yes definitely, writing tests that try to locate and match the rust standard library are brittle. I think in this case you can create a test that just uses one local file (by nesting modules in it).

I agree that the steps in core::complete_from_file could do with being refactored. 
Maybe a function that works out the completion type 'Path' or 'Field' first, and then has separate code to get the context str depending on the type.

Thanks again

yodalee:
I just update the tests file. Now both tests should passed without set `RUST_SRC_PATH`

There is only problem that I cannot set module file name identical in both tests. This will cause one test failed due to `Illegal instruction`. Not sure what causes it.

phildawes:
Hi Yodalee,

Re crashing tests: Rust runs its tests in parallel, so probably the module files are overwriting each other.

Apologies for not merging this earlier.  I've been running with this patch for the last week and thought I had run into a problem with it, but then didn't get round to reproducing the issue. I can't figure out what it was now, seems to work ok as far as I can see.

Thanks again!

PR #565, opened 2016-06-19
fix parsing generic function parameter

### LINKED ISSUE #564: No completion for methods with generic parameters
I'm using racer v1.2.10 

Example:

```rust
struct Foo;

impl Foo {
    fn bar1(&self, x:i32) {}
    fn bar2<T>(&self, x:T) {}
}

fn main() {
    let x = Foo {};
    //  x.   only shows bar1
}
```

yodalee:
The Problem:

The following code is reported having wrong completion.
Both bar1 and bar2 should appear but only bar1 shows
```
struct Foo;

impl Foo {
  fn bar1(&self, x:i32) {}
  fn bar2<T>(&self, x:T) {}
}

fn main() {
  let x = Foo {};
  //  x.   only shows bar1
}
```

The Cause:
In the function nameres.rs:search_scope_for_methods, it will call
`typeinf.rs:first_param_is_self` to check the function belong to Foo. The
function deals with generic parameter like this:
If we find the position of '(' and there is '<' before it, we will skip
this '(' and find next one. The thought is avoiding parenthesis inside
generic parameter, like following case:
pub fn map<U, F: FnOnce(T) -> U>(self, f: F)

However this will skip the parenthesis of arguments if there is no
parenthesis in generic arguments. There is no parenthesis in the remaining
string "&self, x:T)", so function `first_param_is_self` return false

The Solution:
Instead of just skip the parenthesis, we firstly skip the generic arguments
list by skiping the <> pair.
The method is simple: a stack keep record of '<' '> we met. The only
special case is that there may have '->' appear in generic arguments
list.

TODO:
The function `first_param_is_self` now become a little stupid, and the
way it skip generic arguments list may have room for improvement.

Close #564

### DIFF
--- a/src/racer/typeinf.rs
+++ b/src/racer/typeinf.rs
@@ -27,13 +27,24 @@ pub fn generate_skeleton_for_parsing(src: &str) -> String {
 }
 
 pub fn first_param_is_self(mut blob: &str) -> bool {
-    while let Some(start) = blob.find('(') {
-        if blob[..start].contains('<') {
-            // oops - paren in a generic arg
-            // consider 'pub fn map<U, F: FnOnce(T) -> U>(self, f: F)'
-            blob = &blob[start+1..];
-            continue;
+    // skip generic arg
+    // consider 'pub fn map<U, F: FnOnce(T) -> U>(self, f: F)'
+    // we have to match the '>'
+    let mut skip_generic = 0;
+    if let Some(generic_start) = blob.find('<') {
+        blob = &blob[generic_start..];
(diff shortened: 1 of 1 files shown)


yodalee:
Just an thought: why can't we replace function `first_param_is_self` with something like this:
```
blob.find("self")
```
Since `normal` programmer will write `self` as first argument, and there shouldn't have `self` appear in other place.

phildawes:
Thanks @yodalee. You're right, just finding 'self' as an exact ident should do it.

PR #568, opened 2016-06-29
prevent tmp directory remain after run tests

yodalee:
The problem:

After running `cargo test`, there will have one tmpfile directory remain
in racer directory. This is not big but annoying problem to delete them
manually.

The cause:
The directory is generated by testing function finds_nested_submodule_file.
The test will generated directory `tmpfile/sub1/sub2` with one steps, so
when the test is over, only the directory sub2 get removed.

The solution:

Instead of generate sub2 directory in one step. the upmost directory
tmpfile is generated explicitly to allow tmpfile directory being
removed.


phildawes:
Thanks v much @yodalee

PR #570, opened 2016-07-03
Corrected impl lookup to search whole scope and fixed a bug with tuple/unit structs at end of scope

Jonesey13:
This PR fixes two issues. Firstly there is an issue where racer could panic when completing on a tuple or unit struct at the end of a scope:

```
mod sub {
    impl Foo {
        pub fn two(&self) -> usize {2} 
    }
    pub struct Foo(bool);
}

fn main() { // l16
    let t = sub::Foo(true);
    t.t
}
```

Originally, if one tried to complete on `t.t`, racer would panic since it would try to find fields on the struct `Foo`, look ahead for an open/close brace and panic when it meets the `}` at the end of the Sub module. This is because of an `u32` underflow in `scopes::find_close` which has now been fixed.

Secondly, the completion at `t.t` above would still fail since racer would start searching for implementations from the `F` in `pub struct Foo(bool)` rather than from the start of the module scope. This has now been fixed and a new test has been added to cover this case.

### DIFF
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -122,7 +122,8 @@ pub fn search_for_impls(pos: usize, searchstr: &str, filepath: &Path, local: boo
                         session: &Session) -> vec::IntoIter<Match> {
     debug!("search_for_impls {}, {}, {:?}", pos, searchstr, filepath.to_str());
     let s = session.load_file(filepath);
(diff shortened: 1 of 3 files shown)


phildawes:
This is great, thanks @Jonesey13

PR #571, opened 2016-07-04
use repository name instead of package name

xieke91:
I find that racer doesn't work well when I use the git package which the repository name is differnet from the package name.


phildawes:
Hi @xieke91,

Thinks for the patch! Could you give an example of a package that this fixes?

xieke91:
There is a package named [rust-sciter](https://github.com/pravic/rust-sciter) . Racer doesn't work well when I use it.
_**repository name:**_ rust-sciter
_**package name:**_ sciter-rs

PR #576, opened 2016-07-23
Fix Benchmark Tests

Jonesey13:
The benchmarks have been fixed so that they actually compile and run without error (libcollections/bit.rs hasn't existed for quite awhile so the benchmarks now use libcollections/vec.rs). 

The benchmarks have also been moved to the _benches_ folder so that they can be run by _cargo bench_ rather than using feature flags.

I have also removed tests/test.rs as it has been made redundant by the file move.


phildawes:
Cool, thanks @Jonesey13

PR #578, opened 2016-07-24
complete methods from trait in impl of trait

andjo403:
makes it possible to complete and find methods from trait in the impl of that trait.
not sure how to make it work like I want to but I think this is a good first step.

I want to make it complete with the definition of the method but now it only completes the name.
e.g. 
impl fmt::Debug for Match {
   fn fm
          ^ here I want it to complete with "fn fmt(&self, &mut Formatter) -> Result" that is the definition
now it completes with fmt.
Or maybe even better if it figures out that I have no use for the fmt::{Formattter, Result} 
and return "fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result"

### DIFF
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -88,8 +88,46 @@ fn search_scope_for_methods(point: usize, src: Src, searchstr: &str, filepath: &
     let mut out = Vec::new();
     for (blobstart,blobend) in scopesrc.iter_stmts() {
         let blob = &scopesrc[blobstart..blobend];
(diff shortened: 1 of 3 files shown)


andjo403:
one downside with this PR is that due to reuse of search_scope_for_methods I changed so that also the trait declaration is found when trying to complete function calls maybe shall make a new method to find declarations of methods in traits

phildawes:
Hi @andjo403. This is cool, thanks!

Issue #575, opened 2016-07-23
Not able to find some functions


hotpxl:
I'm using Racer to complete `ssh2` crate. My source code looks like this.

```rust
extern crate ssh2;

use ssh2::Session;

fn main() {
    let tcp = TcpStream::connect("127.0.0.1:22").unwrap();
    let mut session = Session::new().unwrap();
    let mut channel = session.channel
}
```

I try to complete directly using `racer` command. I set up `CARGO_HOME` correctly. I hit `racer complete 8 36 <absolute_path_to_file>`(completing right at `session.channel`) and it gives back

```
PREFIX 30,36,channe
MATCH channel_direct_tcpip,380,11,/home/hotpxl/.cargo/registry/src/github.com-1ecc6299db9ec823/ssh2-0.2.12/src/session.rs,Function,pub fn channel_direct_tcpip(&self, host: &str, port: u16,
                                src: Option<(&str, u16)>)
                                -> Result<Channel, Error>
MATCH channel_forward_listen,401,11,/home/hotpxl/.cargo/registry/src/github.com-1ecc6299db9ec823/ssh2-0.2.12/src/session.rs,Function,pub fn channel_forward_listen(&self,
                                  remote_port: u16,
                                  host: Option<&str>,
                                  queue_maxsize: Option<u32>)
                                  -> Result<(Listener, u16), Error>
MATCH channel_open,483,11,/home/hotpxl/.cargo/registry/src/github.com-1ecc6299db9ec823/ssh2-0.2.12/src/session.rs,Function,pub fn channel_open(&self, channel_type: &str,
                        window_size: u32, packet_size: u32,
                        message: Option<&str>) -> Result<Channel, Error>
END
```

But if you go into the `session.rs` file from `ssh2` crate, there is [this other](https://github.com/alexcrichton/ssh2-rs/blob/master/src/session.rs#L362) function that Racer were not able to find. I wonder why it is not showing this function as a candidate.

paholg:
Using racer 1.2.10 with Emacs, I get completions for all four functions:
 
![2016-07-27-001441-s](https://cloud.githubusercontent.com/assets/4908217/17166995/41c0ad3a-538f-11e6-8002-bad3b301e2f2.png)

hotpxl:
On my side I'm using Racer 1.2.10 as well but it doesn't work. I have traced the bug back to PR #577 . Please take a look at the commit, it's not huge, and you'll get it instantly. PR #579 also encountered this issue.

[...]

Issue #555, opened 2016-06-03
No completion for loop variables.


FirstTimeInForever:
`````
struct tr
{
    data: i32
}

fn main()
{
    let mut arr=Vec::<tr>::new();
    arr.push(tr{data: 10});
    arr.push(tr{data: 11});

    for it in arr
    {
        println!("{} ", it.data);
    }
}
`````
After typing `it.` I expect `data` or other members (not only values, also member functions) to be offered but get nothing.

PR #579, opened 2016-07-24
Added basic for loop completion for Vec (IntoIterator)

Jonesey13:
This goes someway towards fixing #555. This PR allows for completion of fields and methods of for loop variables when using a Vec (or, in theory, anything that implements IntoIterator). 

Test example:

```
    struct St
    {
        stfield: i32,
    }
    
    impl St {
        pub fn stmethod(&self) -> u32 {2}
    }
    
    fn main()
    {
        let mut arr: Vec<St> = Vec::new();
        arr.push( St{stfield: 4} );
    
        for it in arr
        {
            it.stf
            it.stm  
        }
    }
```

racer can now complete the field and method from ``it.str``, ``it.stm``.

*What this PR Doesn't Do:* It does not complete on variables from other iterators. The blocker on this is the implementation of IntoIterator for Iterator in libcore/iter/traits.rs:

```
    type Item = I::Item;
    type IntoIter = I;

    fn into_iter(self) -> I {
        self
    }
}
```
This would require
1. racer looking up trait bounds to find potential implementations
2. racer looking outside the trait's defining file to look for implementations 
   (This is because Iterator is defined in libcore/iter/iterator.rs)

This also doesn't deal with issues with racer failing to find the generic types in cases like:
```
let mut v1 = Vec::new();
v1.push("hello".to_string());

let mut v2 = Vec::<String>::new();
```

*Regarding Tests:* The current test setup for racer does not use the Rust source code for travis builds. To get around this issue I have inserted the necessary parts of the source code into a separate &str (like with the Deref test). This hit issue #575, so please merge PR #577 before this.

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -1,6 +1,7 @@
 use core::{self, Match, MatchType, Scope, Ty, Session};
 use typeinf;
 use nameres::{self, resolve_path_with_str};
+use scopes;
 
 use std::path::Path;
 use std::rc::Rc;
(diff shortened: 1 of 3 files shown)


phildawes:
This is great, thanks @Jonesey13

Issue #504, opened 2016-02-06
racer exited with status 101


douglarek:
When jumps to `PI` in `std::f64::consts::PI`; it caused `racer exited with status 101`; see http://is.gd/sA2Ioc.

```
racer v1.2.1
rust 1.6.0
```

thanks

phildawes:
I think racer is struggling with this in libstd/lib.rs
```
#[path = "num/f32.rs"]   pub mod f32;
#[path = "num/f64.rs"]   pub mod f64;
```

phildawes:
(thanks for the issue @douglarek)

m4b:
If @andjo403 's fix works that would be great; I've run into this issue several times now, and I've just used path mods for my library to clean up the user facing API and be as flexible as possible implementation wise; however, racer isn't able to see any of the definitions, which makes it not so great for users.

PR #584, opened 2016-07-31
fix of issue #504

andjo403:
add handling of path attribute when finding mod path.
e.g. #[path = "submod/bar.rs"] pub mod bar;


phildawes:
Cool, thanks @andjo403

PR #588, opened 2016-08-05
Added completion for generic trait impls with one bound

Jonesey13:
The goal of the PR is to allow completion of methods from generic trait implementations (see the new tests for examples). This tackles point 1. brought up in #579, namely finding the into_iter() from:
```
    type Item = I::Item;
    type IntoIter = I;

    fn into_iter(self) -> I {
        self
    }
}
```
This still doesn't tackle issue 2. of finding this implementation in the first place (may have to be a manual file redirect in a follow-up PR).


phildawes:
Great, thanks @Jonesey13

PR #589, opened 2016-08-07
Vomit error message when file cannot be opened

bombless:
When you `racer find-definition` and the target file doesn't exist, racer will print an error message saying slicing index out of range, which is confusing.

I tried to refactor the project using `Result` but hundreds of lines involved, so I figure this band-aid can at least help a little bit for now.


phildawes:
Thanks @bombless

PR #590, opened 2016-08-08
Added basic completion of iter variables in for loops

Jonesey13:
This is a very crude implementation of step 2. from #579. It allows for completion of certain iterators variable in for loops beyond IntoIterator cases. A couple of example cases

Standard iter case for Vec:
```
fn main()
{
    let it: Vec<String> = vec!["Hello".to_string()];
    
    for item in it.iter() {
        item.le               // Will now complete the len() method from String
    }
}
```

Explicit implementation of Iterator:
```
struct St {
    pub item: StItem,
    pub used: bool
}

struct StItem {
    pub field: u32
}
    
impl Iterator for St {
    type Item: StItem;

    fn next(&mut self) -> Option<StItem> {
        if self.used {
            self.used = false;
            return Some(self.item);
        }
        None
    }
}
    
fn main()
{
    let it = St {
        text: StItem { field: 22 },
        used: false
    }
    
    for item in it {
        item.fie          // Will now complete to field from StItem
    }
}
```

*Limitations and workarounds*
* More complex uses of iterators, like zip, currently still don't work due to heavy use of associated types.

* Finding the iter() method on some built in types is near-on impossible atm. For example, for the Vec struct the iter() method is defined as a deref to a slice and the iter() method on slices is defined through the `iterator!()` macro, which would be a very tough challenge to handle (would probably have to expand the macro and store the result in a cache somewhere). I was able to handle the slice deref step but this work ultimately had to be dropped because of the macro issue.
The (hopefully temporary) workaround was to just replace iter() and iter_mut() calls to calls of into_iter(). The return types are the same (modulo references) and doesn't use a macro for implementation on built-in types.

* There is a manual file redirect from iterator.rs to traits.rs so that the following implementation can be located:
```
    type Item = I::Item;
    type IntoIter = I;

    fn into_iter(self) -> I {
        self
    }
}
```

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -460,7 +460,7 @@ impl<'c, 's, 'v> visit::Visitor<'v> for ExprTypeVisitor<'c, 's> {
                         Ty::Match(ref m) =>  {
 
                             match m.mtype {
-                                MatchType::Function => typeinf::get_return_type_of_function(m, self.session)
+                                MatchType::Function => typeinf::get_return_type_of_function(m, m, self.session)
                                     .and_then(|ty| path_to_match(ty, self.session)),
                                 MatchType::Struct => Some(Ty::Match(m.clone())),
                                 _ => {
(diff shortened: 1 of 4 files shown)


phildawes:
Looks great, thanks @Jonesey13

PR #598, opened 2016-08-28
Automatic discovery of source path

### LINKED ISSUE #595: Automatic source discovery with rust's new `rust-src` package
Rust now has an "official" location to install sources to (`<toolchain root>/lib/rustlib/src/rust`). Although there is no separate installer for this new component, it can be installed via rustup.

edit: scratch everything that was previously here, I'm an idiot :disappointed: 

It's as simple as `racer` calling `rustc --print sysroot` and locating the source that way. This will then seamlessly work with rustup.

Diggsey:
The environment variable still has higher priority. For editor integrations where the environment variable cannot be unset, setting it to an empty string will have the same effect as unsetting it.

Fixes #595

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -27,6 +27,8 @@ use std::path::{Path, PathBuf};
 use std::io::{self, BufRead};
 #[cfg(not(test))]
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
+#[cfg(not(test))]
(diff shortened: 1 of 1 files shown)


phildawes:
Thanks @Diggsey this looks good. We'll need to keep an eye on whether invoking rustc causes noticeable additional overhead for non-daemon users.

PR #604, opened 2016-09-03
Remove README note about rustup.rs

birkenfeld:
The code in question has been removed in 7455a4841e8c6d588a669329705fd8a42c230bea
so the note should go.


jwilm:
Sounds right; thanks for the patch!

PR #605, opened 2016-09-03
Removed Rustup-specific configuration

davidbarsky:
As per discussion [here](https://www.reddit.com/r/rust/comments/50w0q2/racer_not_autocompleting_on_with_third_party/) and testing on my local machine, Racer no longer requires setting of `CARGO_HOME`.

An alternative to removing Rustup-specific configuration is to add a note letting users that they no longer need to configure their environment specifically for Rustup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,17 +20,17 @@ As mentioned in the command output, don't forget to add the installation directo
 
 ### From sources
 
-1. Clone the repository: ```git clone https://github.com/phildawes/racer.git```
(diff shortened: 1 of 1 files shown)


jwilm:
Is this the same as #604?

davidbarsky:
...it most certainly is. i forgot to check other open pull requests, my bad!






On Sat, Sep 3, 2016 at 1:01 PM -0400, "Joe Wilm" <notifications@github.com> wrote:












Is this the same as #604?

PR #607, opened 2016-09-04
Several improvements to main and tests

birkenfeld:
A few refactorings at first, then lots of housekeeping and minor performance improvements.


jwilm:
Merging a bunch of the other PRs seems to have created a conflict here. Would you mind updating your branch?

src/racer/core.rs
```diff
@@ -286,14 +286,15 @@ impl fmt::Debug for PathSearch {
...
+               self.filepath.display(),
                self.point)
     }
 }
 
 pub struct IndexedSource {
     pub code: String,
-    pub idx: Vec<(usize, usize)>
+    pub idx: Vec<(usize, usize)>,
+    pub lines: RefCell<Vec<(usize, usize)>>
```
jwilm: Could this just be initialized in the constructor so a `RefCell` isn't needed?

birkenfeld: I wanted to initialize the indices lazily because they're probably not used for many files.

But it could conceivably be done in the same iteration that generates the statement indices.

jwilm: Seems fine then.

jwilm:
Is there more you are hoping to do with this branch before having it merged? I've looked over what's here and it all looks great. Maybe if there's additional changes you could open another PR?

birkenfeld:
I think it can be merged for now. I don't know when I'll be able to work on Racer again. Thanks for the review!

jwilm:
Thanks @birkenfeld

PR #609, opened 2016-09-04
Should now find docs for methods (fixes #597).

### LINKED ISSUE #597: Racer does not return docstrings in some impl functions
```
$ cat example.rs
fn main() {
    let v: Vec<_> = vec![];
    v.remove(1);
}
$ racer complete-with-snippet 3 9 example.rs example.rs                    
PREFIX 6,9,rem
MATCH remove;remove(${1:index});631;11;/home/wilfred/src/rustc-1.10.0/src/libstd/../libcollections/vec.rs;Function;pub fn remove(&mut self, index: usize) -> T;""
END
```

The location racer suggests does indeed have a docstring:

```rust
    /// Removes and returns the element at position `index` within the vector,
    /// shifting all elements after it to the left.
    ///
    /// # Panics
    ///
    /// Panics if `index` is out of bounds.
    ///
    /// # Examples
    ///
    /// ```
    /// let mut v = vec![1, 2, 3];
    /// assert_eq!(v.remove(1), 2);
    /// assert_eq!(v, [1, 3]);
    /// ```
    #[stable(feature = "rust1", since = "1.0.0")]
    pub fn remove(&mut self, index: usize) -> T {
```

so I'm puzzled that racer doesn't include this docstring.

birkenfeld:


jwilm:
Thanks!

Wilfred:
Thank you! :)

PR #623, opened 2016-09-27
canonicalize path to make test pass in windows

andjo403:
the canonicalize add \\?\ at the begining of the path in windows so to make the test cargo::tests::get_crate_file_from_overrides pass in windows canonicalize needs to be on actual and expected.


jwilm:
Oh! this is great. Thank you for submitting a patch. I'll add an issue to get tests running on appveyor so this doesn't happen again.

Issue #628, opened 2016-10-01
Character literals containing more than one source character are incorrectly masked into invalid syntax


Arnavion:
- rust 1.12.0 (3191fbae9 2016-09-23) (via rustup, stable-x86_64-pc-windows-msvc toolchain)
- racer 1.2.10 (latest via `cargo install racer`), also [current git master](https://github.com/phildawes/racer/commit/b36a9a6af862a63db9397e76abc3cb2013751b8d)

`C:\Users\Arnavion\Desktop\test.rs`

``` rust
fn main() {
    let mut splits = "555".split_terminator('\n');
    splits.pop()
}
```

`SET RUST_LOG=racer=debug racer find-definition 3 12 C:\Users\Arnavion\Desktop\test.rs`

(`(3, 12)` is in the middle of `pop`)

This panics with:

```
DEBUG:racer::core: find_definition_ for |"splits"| |"pop"| Field
DEBUG:racer::ast: visit_expr expr(4294967295: splits)
DEBUG:racer::ast: expr is a path P[splits]
DEBUG:racer::ast: resolve_ast_path P[splits]
DEBUG:racer::nameres: resolve_path_with_str P[splits]
DEBUG:racer::nameres: resolve_path P[splits] Some("C:\\Users\\Arnavion\\Desktop\\test.rs") 75 ExactMatch
DEBUG:racer::nameres: resolve_name splits Some("C:\\Users\\Arnavion\\Desktop\\test.rs") 75 ExactMatch BothNamespaces
DEBUG:racer::nameres: search_local_scopes PathSegment { name: "splits", types: [] } Some("C:\\Users\\Arnavion\\Desktop\\test.rs") 75 ExactMatch BothNamespaces
DEBUG:racer::nameres: searching scope BothNamespaces start: 11 point: 75 'splits' Some("C:\\Users\\Arnavion\\Desktop\\test.rs") ExactMatch local: true, session: Session("C:\\Users\\Arnavion\\Desktop\\test.rs", "C:\\Users\\Arnavion\\Desktop\\test.rs")
DEBUG:racer::matchers: match_pattern_let point is 24
DEBUG:racer::typeinf: get_type_of match Match ["splits", Some("C:\\Users\\Arnavion\\Desktop\\test.rs"), 24, true, Let, [], [] |let mut splits = "555".split_terminator('\n');|]
DEBUG:racer::typeinf: get_type_of_let_expr calling parse_let |let mut splits = "   ".split_terminator('  ');|
[...]

PR #630, opened 2016-10-01
Add tests for #628.

Arnavion:
First test passes, second test fails in the same way as #628

```
error: character literal may only contain one codepoint: '
 --> bogofile:1:22
  |
1 | let foo = Foo { bar: '  ', }
  |                      ^^

test works_with_character_literals_containing_more_than_one_character ... FAILED

failures:

---- works_with_character_literals_containing_more_than_one_character stdout ----
        thread 'works_with_character_literals_containing_more_than_one_character' panicked at 'Box<Any>', C:\Users\Arnavion\.cargo\registry\src\github.com-1ecc6299db9ec823\syntex_syntax-0.43.0\src\parse\lexer/mod.rs:1279
```

### DIFF
--- a/tests/system.rs
+++ b/tests/system.rs
@@ -1977,6 +1977,48 @@ fn finds_self_param_when_fn_has_generic_closure_arg() {
 }
 
 #[test]
+fn works_with_character_literals_containing_one_character() {
+    // issue #628
+    let src = "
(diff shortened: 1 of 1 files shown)


kngwyu:
Closed since I can't reproduce this bug in current version.

Issue #619, opened 2016-09-20
Versioning


jwilm:
I was about to make a cut of Racer this morning and realized that we've changed a public API in an incompatible way. Specifically, we landed #498 which removed some unsafe code related to the `FileCache`. This was a very welcome patch of course, but it means we can't release from master without bumping the major version.

Options:
1. **Revert the FileCache change**. This is probably no longer easy to do since there's been quite a few changes to master.
2. **Actually cut a 2.0**. This isn't exactly great since there's a lot of other API work that could be done, but it would allow us to move forward with minimal effort.
3. **Add a FileCache shim**. This could expose an API matching the old FileCache, but it's pure technical debt. The shim would be thrown out when we actually

I'm inclined to just cut a version 2.0. This is the simplest option, there's just one type with breaking changes (so upgrading is easy), and there's really no limit to how many major versions we can have.

There's a bunch of fixes on master that really ought to be released, and there's more on the way. Assuming we release a 2.0 now, I would want to figure out what a minimal API we can commit to for 3.0 would look like and hold off on breaking changes until that's decided. Until then, the focus would be on backwards compatible changes like increasing stability and making racer work for more cases.

@phildawes, thoughts?

phildawes:
Hello! Does anything other than racerd use this api? I think the other plugins just use the command line interface. If that's true, then probably best to cut a 1.3 and not worry about it too much.
(I'm happy for a 2.0 release if you think it's warranted though)

jwilm:
Interesting thought! I think most integrations are using either the CLI, stdio pipe, or racerd. Thanks, Phil. 

@nrc, does RLS use the FileCache API?

nrc:
Not really. Our entry point to Racer is `complete_from_file`. We do create an empty `FileCache`, but only to make a `Session` which we pass to `complete_from_file`, we don't use any functionality beyond that.

jwilm:
@nrc thanks for the input

After thinking on this a bit (and reading comments in [futures-rs#160]), I think we should go for a 2.0 just to avoid any unexpected breakage. With that, I'd like to reduce the public API to a minimal set of things needed to make completions and etc; hopefully doing so will make this less likely to happen again.

[futures-rs#160]: https://github.com/alexcrichton/futures-rs/issues/160

jwilm:
[...]

PR #632, opened 2016-10-03
Initial v2 of the racer library API

jwilm:
Version 2.0 of the Racer API is introduced in order to provide
compatibility guarantees to dependents on crates.io. This is
accomplished by exposing only a very narrow but powerful API capable of
performing find_definition, find_completions, and etc. as was possible
before. The command-line API remains unchanged.
## Motivation

See #619 for some reasoning. The short version is: we want to provide a simple API which doesn't expose internals so we can iterate quickly without breaking the public API. There's a bunch of changes on master that haven't been released on crates.io due to versioning issues, and this is intended to prevent such situations moving forward.
## Major changes
- All modules are now private with only bare necessities exposed
  publicly
- All publicly available functions, types, and structs are now exposed
  directly under `racer` instead of `racer::core`
- `Session::with_path` has been removed, and `Session::new` has replaced
  it. The `with_path` variant was an additional and confusing way to
  support substitute files. Substitute files can still be used by
  preloading the contents into the `Session` using the query file's
  path. All requests for the query file's source will return the
  substitute file contents.
- `FileCache` is now an opaque blob; use the `Session` to manipulate it
## TODO (more major changes)

These items are planned changes for 2.0 but are not yet completed
- [x] Update the find_definition and completion APIs to take a `Cursor` enum
  which is either `Cursor::ByteOffset(usize)` or
  `Cursor::Coordinate { line: usize, col: usize }`. This will allow the
  `Src` and `IndexedSource` types to remain private and will allow
  some methods on `Session` to become private.
- [x] Return a `Completion`, `Definition`, or some unified type from
  corresponding functions. Basically, we expose a `vec::IntoIter<Match>`
  which is an oddly specific type. Instead, we should expose some racer
  type which is iterable (for when multiple values are returned) with
  `Item=Match`, and in the case of completions, contains the prefix for
  which matches were found. The goal here is to hide more internals and
  provide an API with only info relevant to consumers
- [x] Build prefix detection into racer instead of calling internals from
  the racer/main.rs source.
- [x] Add a `Loader` trait used by `FileCache` for RLS.
- [x] Update racerd
- [x] Update RLS
## Doc screenshots

I wanted to provide a preview of the new API without too much effort or risk of google indexing it, so here's a few screenshots. Stuff that is still highly subject to change is not included.
### Racer root

This is subject to changes listed in **TODO** section. Note that no modules are exposed: this is the entirety of the Racer API.

image
### Session API

This should be pretty close to the final `Session` API

image
### FileCache

No screenshot because it _only_ has a `new()` method. It's now an opaque blob.
## Feedback

Would really enjoy some feedback on the direction here from @phildawes @nrc @jonathandturner and anyone else who use consuming Racer via the Rust API.


nrc:
Sorry I didn't look through the code to see if this comment is reasonable, but I'm not sure I understand Racer's API well enough, even if I did (@jntrnr has been doing most of the Racer integration in the RLS).

From our perspective, a key requirement is to skip working with Racer's FileCache and work use our own - we need to use our own cache for a whole tonne of other stuff. I don't mind pusing our in-memory content into Racer's cache if that makes it easier for Racer to work with, but we have to absolutely avoid being restricted to reading files from disk (or really ever having a file concept if possible). Ideally though we would skip Racer's cache (because I assume it must clone the text it caches) and we could pass an `&str` to Racer and Racer would operate on it (I guess that is probably not possible since it might need to open other files though).

One possibility is you could the FileCache into a trait and we could implement it for our cache? Would that work?

jwilm:
@nrc Yeah that's totally possible. Can you share the API your own cache uses?

nrc:
It is here: https://github.com/nrc/rls-vfs/blob/master/src/lib.rs but it's likely to change a lot

jwilm:
I think the thing to do might be allowing you to provide some `FileLoader` which the file cache would use internally when it has a miss. Although.. this would have a problem where there's extra copies of the file floating around (ie our cache and RLS). Maybe not a big deal. If we try and go for a more memory efficient version, things would need to change in a more dramatic way within Racer. 

How do you feel about passing a `FileLoader` and maybe optimizing this later?

nrc:
We'd have to be able to invalidate Racer's cache too, otherwise you'd hardly ever hit our loader. Otherwise this sounds ok - correct but inefficient.

jwilm:
> We'd have to be able to invalidate Racer's cache too, otherwise you'd hardly ever hit our loader

That's a very good point. I'll see what I can come up with this weekend.

jwilm:
@nrc sorry for the delay here; I'm making time this week to finish this up. Everything is done now except the `FileLoader` for _RLS_ which I'm hoping to finish tomorrow morning.

jwilm:
Btw I've [uploaded some docs](https://pushdev.jwilm.io/racer/racer/) in case you want to look around

jwilm:
@nrc 

Updated the PR with an initial version of a [`FileLoader`] for the cache. Additionally, there's a [`remove_file()`](https://pushdev.jwilm.io/racer/racer/struct.FileCache.html#method.remove_file) method on the cache for invalidating old files. To prevent an explosion of type parameters throughout Racer, the [`FileLoader`] is boxed up in [`FileCache::new()`]; this unfortunately adds a `'static` bound. Let me know if that's an issue and I'll figure something else out. Would love feedback either way :smile: 

[`FileCache::new()`]: https://pushdev.jwilm.io/racer/racer/struct.FileCache.html#method.new
[`FileLoader`]: https://pushdev.jwilm.io/racer/racer/trait.FileLoader.html

nrc:
So with `remove_file` the intent is that the client calls it to invalidate a file in Racer's cache?

I've only looked closely at the FileCache stuff, and that looks good to me.

jwilm:
Based on the conversation in https://github.com/jonathandturner/rustls/pull/67, there's 3 additional *TODO*s:

1. Instead of taking a `Cursor`, methods should take `Into<Cursor>`.
2. Provide implementations of `From<usize> for Cursor` and `From<Coordinate> for Cursor`
3. Rename `Cursor` to `Location`

jwilm:
Going to hold off on adding `Result` to return types for now since it's likely impossible to incrementally add in a backwards compatible way. We can incrementally use `Result` internally and just map it to empty iterators and `None` at the public API boundary. At some point we can actually expose them and cut a v3.

Everything else on the list is done, so this should be good to go!

PR #636, opened 2016-10-18
add support to get concrete type of method with return type Self

### LINKED ISSUE #587: No completion for struct
```
extern crate byteorder;

use std::io;
use std::io::prelude::*;
use std::io::SeekFrom;
use byteorder::{ReadBytesExt, /*WriteBytesExt,*/ BigEndian, LittleEndian};

#[derive(Debug)]
pub struct MyStruc<T> {
    _io : T,
    pos : u64
}

impl<T : Read + Write> MyStruc<T> {
 pub fn new(io: T) -> Self {
     let me = MyStruc {
        _io: io,
        pos: 0,
     };
     me
 }

// omitted other functions
 pub fn read_value(&mut self) -> io::Result<u64> {
    let c = try!(self._io.read_u64::<LittleEndian>());
    self.pos += 8;
    Ok(c)
 }
}


fn some_func() {
     use std::path::Path;
     use std::fs::File;

     if let Ok(f) = File::open(Path::new("./src/testfile")) {
         let mut x = MyStruc::new(&f); //this gets completed
         let y = x.read_value().expect("Failed to read value"); // x. does not
     }
}
```

andjo403:
fixes #587

### DIFF
--- a/src/racer/typeinf.rs
+++ b/src/racer/typeinf.rs
@@ -334,6 +334,15 @@ pub fn get_return_type_of_function(fnmatch: &Match, contextm: &Match, session: &
         ast::parse_fn_output(decl, Scope::from_match(fnmatch))
     });
 
+    // Convert output arg of type Self to the correct type
+    if let Some(core::Ty::PathSearch(ref mut path, _)) = out {
+        if let Some(ref mut path_seg) = path.clone().segments.get(0) {
+            if "Self" == &path_seg.name {
+                return get_type_of_self_arg(fnmatch, src.as_src(), session);
+            }
+        }
+    }
+
     // Convert a generic output arg to the correct type
     if let Some(core::Ty::PathSearch(ref mut path, _)) = out {
         if let Some(ref mut path_seg) = path.clone().segments.get(0) {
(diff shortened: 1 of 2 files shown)


src/racer/typeinf.rs
```diff
@@ -334,6 +334,15 @@ pub fn get_return_type_of_function(fnmatch: &Match, contextm: &Match, session: &
         ast::parse_fn_output(decl, Scope::from_match(fnmatch))
     });
 
+    // Convert output arg of type Self to the correct type
+    if let Some(core::Ty::PathSearch(ref mut path, _)) = out {
+        if let Some(ref mut path_seg) = path.clone().segments.get(0) {
```
jwilm: Is this clone necessary?

andjo403: I'm still not that good at rust so I copied the code from some lines below.
but do now not know why I have "ref mut" either 
is this better?
`    if let Some(core::Ty::PathSearch(ref path, _)) = out {
        if let Some(ref path_seg) = path.segments.get(0) {
            if "Self" == path_seg.name {
`
I can remove the ref in Some with all test passing also but do not know what is best when I tried to remove the ref in the first stamtent the borrow checker complained.

jwilm: Ah yeah. Without the `ref` in `if let Some(ref thing) = ...`, it's expecting to take whatever's on the rhs instead of borrowing it. `ref` there basically says you want to borrow the value temporarily. If the code compiles without the `clone()`, you're good to go!

andjo403:
thank you for the fast reviews

jwilm:
Thanks for the feature!

PR #649, opened 2016-11-11
Search workspace for lockfile, if it exists

vvuk:
If there's a workspace, search it for a lockfile instead of the same dir as Cargo.toml.

### DIFF
--- a/src/racer/cargo.rs
+++ b/src/racer/cargo.rs
@@ -639,9 +639,22 @@ pub fn get_crate_file(kratename: &str, from_path: &Path) -> Option<PathBuf> {
 
     if let Some(tomlfile) = find_cargo_tomlfile(from_path.to_path_buf()) {
         // look in the lockfile first, if there is one
+        // check if there's a workspace first.
         trace!("get_crate_file tomlfile is {:?}", tomlfile);
+        let mut workspace: Option<String> = None;
+        if let Some(toml) = parse_toml_file(&tomlfile) {
+            if let Some(package) = toml.get("package") {
+                if let Some(w) = package.lookup("workspace") {
+                    workspace = Some(w.as_str().unwrap().to_owned());
+                }
+            }
+        }
         let mut lockfile = tomlfile.clone();
         lockfile.pop();
+        if workspace.is_some() {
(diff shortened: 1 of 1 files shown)


Dushistov:
May be related to #687

src/racer/cargo.rs
```diff
@@ -639,9 +639,22 @@ pub fn get_crate_file(kratename: &str, from_path: &Path) -> Option<PathBuf> {
 
     if let Some(tomlfile) = find_cargo_tomlfile(from_path.to_path_buf()) {
         // look in the lockfile first, if there is one
+        // check if there's a workspace first.
         trace!("get_crate_file tomlfile is {:?}", tomlfile);
+        let mut workspace: Option<String> = None;
+        if let Some(toml) = parse_toml_file(&tomlfile) {
```
jwilm: This block can be simplified a lot with combinators:

```rust
let workspace = parse_toml_file(&tomlfile)
    .and_then(|toml| toml.get("package"))
    .and_then(|package| package.lookup("workspace"))
    .map(|workspace| workspace.as_str().unwrap().to_owned());
```

jwilm: Oh, you might need to save off the parsed toml file to avoid borrowck issues:

```rust
let toml = parse_toml_file(&tomlfile);
let workspace = toml.and_then( // .. as before
```

danburkert:
This PR only works if the member crate in the workspace includes the `package.workspace` key in it's `Cargo.toml`, which per the [Cargo docs](http://doc.crates.io/manifest.html#the-workspace--field-optional) is optional.  I believe the correct thing to do is recursively search up the directory tree if the key doesn't exist.

GoldenBadger:
Hi, will there be any progress on this PR? Currently racer does not work if your project involves a workspace with multiple sub-crates.

TedDriggs:
@GoldenBadger I don't think @vvuk is actively working on this any more. We'd welcome a new PR.

GoldenBadger:
Hi, I've submitted my PR at #792.

TedDriggs:
Closing in favor of @GoldenBadger's work

PR #650, opened 2016-11-12
Handle trailing slash in RUST_SRC_PATH

solidsnack:
Presently, if a developer unwittingly includes the slash, they get a confusing error message (`racer can't find the directory pointed to by the RUST_SRC_PATH variable ""`) and it seems perfectly harmless if they do include it.


TedDriggs:
Closing per @birkenfeld's guidance

PR #656, opened 2016-12-07
Remove println from find_definition

### LINKED ISSUE #655: Racer is writing some output to stderr in `find_definition`
I think it is the identifier being searched for. This is interfering with the RLS's output.

jwilm:
Resolves #655 

Couldn't find anything that wrote to stderr, but this println sounded like it might be the problem.

cc @nrc


jwilm:
Adding the clippy lint actually detected another println! I had missed. Will merge as soon as travis passes.

Issue #657, opened 2016-12-07
Add clippy test on travis


jwilm:
Among other things, there's a lint for catching `println!` statements outside of test code which should prevent regressions like to #655 in RLS.

jwilm:
Resolved

PR #658, opened 2016-12-07
Run clippy on travis

jwilm:
Currently only the print_stdout lint is denied as part of preventing
issues like #655.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,15 @@
 language: rust
 
+rust:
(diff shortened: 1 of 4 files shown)


jwilm:
Resolves #657

jwilm:
Travis shows failing due to the new clippy lint. There are no other issues with the tests. I'm merging this as-is to reflect the broken state of master. It should go back to green after #655 is resolved.

PR #659, opened 2016-12-08
Update dependencies

pmarcelll:


jwilm:
Wow; thanks! I'm surprised 9 versions of syntex didn't require more changes. Thanks for taking the time to put this together!

pmarcelll:
Yeah, it surprised me, too. I guess syntex is slowing down/maturing due to not many new language changes. Maybe when stable `impl Trait`, `const fn`, type level integers, non-lexical lifetimes, HKT, etc. comes...
And thanks for maintaining `racer`!

PR #660, opened 2016-12-11
validate rust_src path later to fix #643

### LINKED ISSUE #643: Racer help commands should not require RUST_SRC_PATH
```
~ $ racer --help
RUST_SRC_PATH environment variable must be set to point to the src directory of a rust checkout. E.g. "/home/foouser/src/rust/src"
~ $ racer --version
RUST_SRC_PATH environment variable must be set to point to the src directory of a rust checkout. E.g. "/home/foouser/src/rust/src"
```

Ideally users could run these commands without setting these environment variables.

andjo403:


jwilm:
Thanks!

PR #662, opened 2016-12-22
use repository name instead of package name

xieke91:
fix bug that racer can't find the right source directory when use branch key in cargo.toml.


jwilm:
Thanks!

PR #665, opened 2016-12-23
Handle `self` after use path

sinkuu:


jwilm:
This looks great; thanks!

Issue #641, opened 2016-11-02
Autocomplete does not work through type aliases


TedDriggs:
Static and instance methods are not completed for type aliases. Example below:

```rust
#[derive(Debug, Clone, Eq, PartialEq)]
/// An expression parameter
pub struct Param<T : Eq + Hash> {
    allowed: HashSet<T>
}

impl<T : Eq + Hash> Param<T> {
    pub fn new<V: Iterator<Item=U>, U: Into<T>>(allows: V) -> Param<T> {
        let mut ep = Param { allowed: HashSet::new() };
        for i in allows {
            ep.allowed.insert(i.into());
        }
        
        // return the constructed ExprParam
        ep
    }
    
    pub fn accepts<U>(&self, v: U) -> bool where T : PartialEq<U> {
        self.allowed.iter().any(|i| *i == v)
    }
}

// separate file in the same folder
pub type ExprParam = Param<ExprParamType>;
```

runcom:
fixed by #666?

TedDriggs:
Yes, it seems to be working as of 2.0.8.

PR #666, opened 2016-12-23
Fix type alias resolution

sinkuu:
`get_type_of_typedef` was looking for outer scope even if it's a type alias, causing overflow (`0 - 1`) or missing items.


sinkuu:
(This doesn't fix #641)

jwilm:
This is great; thanks :D

Issue #654, opened 2016-12-06
Release 2.0 on crates.io


jwilm:
The 2.0 API was merged into master recently, but it hasn't been published to crates.io. This issue is a reminder to do so and a pre-publication checklist.

- [ ] Land remaining open PRs
- [ ] Update changelog
- [ ] Publish

PR #667, opened 2016-12-23
add suggestion for git based cargo install

waynenilsen:
I wanted to do this for compatibility with `rustup` before it was released to crates.io

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,13 @@
 
 ### With `cargo install`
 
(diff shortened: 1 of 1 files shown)


jwilm:
I think I'm going to decline this and actually get v2 published, and then the cargo install instructions should work again. cc #654

jwilm:
Thanks for the PR, though :D

jwilm:
2.0 Published

PR #669, opened 2016-12-25
Fix `use foo::bar::self;` not followed

sinkuu:
`use foo::bar::self` is parsed as `ViewPathSimple(self, foo::bar)`.


sinkuu:
Also fixed a bug reported [here](https://www.reddit.com/r/rust/comments/5km2ly/racer_doesnt_get_all_completions/).

jwilm:
Thanks for the fix. I'll get a release out shortly.

PR #675, opened 2016-12-31
fixes #618 and fixes #594

### LINKED ISSUE #594: complete-with-snippet does not always include full docstring
```
$ cat foo.rs
/// Hello world
/// (quux)
pub fn foo() {}

pub fn bar() {
    foo();
}
$ racer complete-with-snippet 6 7 foo.rs foo.rs
PREFIX 4,7,foo
MATCH foo;foo();3;7;foo.rs;Function;pub fn foo();"Hello world"
END
```

I'm not sure why, but something about parentheses on the last line of the docstring seems to prevent racer from including that line.

### LINKED ISSUE #618: Missing doc line on OsRng
When trying to get the documentation for a the `OsRng` struct from the `rand` crate, the last line is missing. Example:
Here's the documentation of the `OsRng` struct:

```
/// A random number generator that retrieves randomness straight from
/// the operating system. Platform sources:
///
/// - Unix-like systems (Linux, Android, Mac OSX): read directly from
///   `/dev/urandom`, or from `getrandom(2)` system call if available.
/// - Windows: calls `CryptGenRandom`, using the default cryptographic
///   service provider with the `PROV_RSA_FULL` type.
/// - iOS: calls SecRandomCopyBytes as /dev/(u)random is sandboxed.
/// - PNaCl: calls into the `nacl-irt-random-0.1` IRT interface.
/// This does not block.
```

Here's what I get from `racer complete-with-snippet 28 15 src/main.rs` (28:15 being at the end of `rand::OsRng`):

```
PREFIX 10,15,OsRng
MATCH OsRng;OsRng;27;11;/home/timo/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.3.14/src/os.rs;Struct;pub struct OsRng(imp::OsRng)\;;"A random number generator that retrieves randomness straight from\nthe operating system. Platform sources:\n\n- Unix-like systems (Linux, Android, Mac OSX): read directly from\n  `/dev/urandom`, or from `getrandom(2)` system call if available.\n- Windows: calls `CryptGenRandom`, using the default cryptographic\n  service provider with the `PROV_RSA_FULL` type.\n- iOS: calls SecRandomCopyBytes as /dev/(u)random is sandboxed.\n- PNaCl: calls into the `nacl-irt-random-0.1` IRT interface."
END
```

So the last line is missing.

If I copy the exact same documentation to replace the documentation of another struct (I tried it with `IsaacRng`) however, it is correctly displayed for that struct.

Edit:
Minimal example to reproduce:
`cargo new --bin minimal`
Add dependency `rand = "0.3"` to Cargo.toml.
Modify src/main.rs:

```
extern crate rand;
fn main() {
    rand::OsRng
}
```

`racer complete-with-snippet 3 15 src/main.rs`

andjo403:
when the find_doc is called with a point at the line before the match the last line in the doc is missing
e.g. when fn declaration is at begining of line.
Use the match point instead to make sure the correct line is used.

also allow empty lines between doc comment and match line.

### DIFF
--- a/src/racer/matchers.rs
+++ b/src/racer/matchers.rs
@@ -391,7 +391,7 @@ pub fn match_struct(msrc: &str, blobstart: usize, blobend: usize,
             contextstr: get_context(blob, "{"),
             generic_args: generics.generic_args.into_iter().map(|arg| {arg.name}).collect(),
             generic_types: Vec::new(),
-            docs: find_doc(msrc, blobstart),
+            docs: find_doc(msrc, blobstart + start),
         })
     } else {
         None
@@ -418,7 +418,7 @@ pub fn match_type(msrc: &str, blobstart: usize, blobend: usize,
             contextstr: first_line(blob),
             generic_args: Vec::new(),
             generic_types: Vec::new(),
-            docs: find_doc(msrc, blobstart),
+            docs: find_doc(msrc, blobstart + start),
         })
     } else {
         None
@@ -445,7 +445,7 @@ pub fn match_trait(msrc: &str, blobstart: usize, blobend: usize,
             contextstr: get_context(blob, "{"),
             generic_args: Vec::new(),
             generic_types: Vec::new(),
-            docs: find_doc(msrc, blobstart),
+            docs: find_doc(msrc, blobstart + start),
         })
     } else {
         None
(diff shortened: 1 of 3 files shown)


jwilm:
Looks good; thanks!

PR #684, opened 2017-01-12
Fix #683: do not ignore empty module doc lines

### LINKED ISSUE #683: Consecutive newline characters missing in docstring
## Description
Racer seems to either remove consecutive newline characters or ignore (inner) empty lines, thus breaking markdown rendering in certain places, for example `std::path`.

### The rust source-code:
```rust
//! * The root directory component represents a separator that designates
//!   starting from root. For example, `/a/b` has a root directory component
//!   followed by normal components `a` and `b`.
//!
//! On Windows, an additional component type comes into play:
//!
//! * Prefix components, of which there is a large variety. For example, `C:`
//! and `\\server\share` are prefixes. The path `C:windows` has a prefix
//! component `C:` and a normal component `windows`; the path `C:\windows` has a
//! prefix component `C:`, a root directory component, and a normal component
//! `windows`.
//!
//! ### Normalization
//!
//! Aside from splitting on the separator(s), there is a small amount of
//! "normalization":
```

### The returned doc-string is (excerpt with un-escaped special characters):
```markdown
* The root directory component represents a separator that designates
  starting from root. For example, `/a/b` has a root directory component
  followed by normal components `a` and `b`.
On Windows, an additional component type comes into play:
* Prefix components, of which there is a large variety. For example, `C:`
and `\\server\share` are prefixes. The path `C:windows` has a prefix
component `C:` and a normal component `windows`; the path `C:\windows` has a
prefix component `C:`, a root directory component, and a normal component
`windows`.
### Normalization
Aside from splitting on the separator(s), there is a small amount of
"normalization":
```

In this example, the two separated lists get merged due to a missing newline character.

qzed:
## Problem
Racer expected an empty module documentation line to start with <code>//! </code>, including a space. The Rust documentation does not include this space (at least not in `std::path`) , thus racer ignores all empty lines of the documentation, resulting in markdown being rendered differently in some cases (see #683).

## Solution
This PR modifies the pattern to `//!`, excluding the space, to fix this.

## Possible Issues
None that I am aware of.

### DIFF
--- a/src/racer/matchers.rs
+++ b/src/racer/matchers.rs
@@ -723,7 +723,7 @@ fn find_mod_doc(msrc: &str, blobstart: usize) -> String {
         // Skip over the copyright notice and empty lines until you find
         // the module's documentation (it will go until the end of the
         // file if the module doesn't have any docs).
-        .filter(|line| line.starts_with("//! "))
+        .filter(|line| line.starts_with("//!"))
         .peekable();
 
     // Use a loop to avoid unnecessary collect and String allocation
--- a/tests/system.rs
+++ b/tests/system.rs
@@ -1062,6 +1062,33 @@ fn finds_external_fn_docs() {
 }
 
 #[test]
+fn keeps_newlines_in_external_mod_doc() {
+    let _lock = sync!();
+
+    // issue 683: do not remove newlines inside of mod-doc
+    let src1 = "// Copyright notice
+
+//! The mods multiline documentation
+//!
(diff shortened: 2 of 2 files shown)


TedDriggs:
@qzed thanks for submitting this PR! Would you mind adding a test to `tests/system.rs` which shows this working correctly? That way we'll avoid regressions in the future.

qzed:
No problem, I've added a test.

PR #686, opened 2017-01-16
handle whitespace and comments in call chain fix #481

### LINKED ISSUE #481: Racer could autocomplete chained calls
Racer completes usual function calls like:

``` rust
use curl::http::handle()
```

(`handle()` will be autocompleted)
but doesn't analyze returned value:

``` rust
use curl::http::handle().post(uri, body)
```

(`post()` call won't be autocompleted)
It is advanced feature but it would be definitely useful.
edit: can be related to https://github.com/phildawes/racer/issues/443

andjo403:
#468 can also be closed

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -898,7 +898,7 @@ fn complete_from_file_(
     cursor: Location,
     session: &Session
 ) -> vec::IntoIter<Match> {
-    let src = session.load_file(filepath);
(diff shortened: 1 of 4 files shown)


jwilm:
This looks great; thanks!

jwilm:
Published a new version with this - 2.0.5

PR #689, opened 2017-01-25
WIP: Make FileCache being Send and Sync

jonasbb:
This pull request is just to be able to discuss the implementation and how it works together with RLS.
I used `unwrap`s because there is no easy way to avoid them without a lot of re-factoring.

* Make `FileLoader` require being `Send`
* Convert `Rc` to `Arc`
* Convert `RefCell<_>` in `IndexedSource` to `Arc<RwLock<_>>` to make `FileCache`
  Sync

This is necessary to implement jonathandturner/rls#33 which requires
sharing a `FileCache` between different threads.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ toml = "~0.2.1"
 env_logger = "~0.3.4"
 typed-arena = "~1.2.0"
 clap = "~2.19.0"
+parking_lot = "~0.3.6"
 
 [dependencies.clippy]
 version = "0.0.103"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,6 +7,7 @@ dependencies = [
  "env_logger 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "parking_lot 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -28,6 +28,7 @@ toml = "~0.2.1"
 env_logger = "~0.3.4"
 typed-arena = "~1.2.0"
 clap = "~2.19.0"
+parking_lot = "~0.3.6"
```
nrc: I would prefer using the std lib primitives since they support poisoning. While parking lot is probably a better implementation, I fear that having unusual (c.f., standard Rust) semantics in edge cases is asking for trouble.

jwilm: Not having to deal with possible poison errors here is kind of nice, and the parking_lot RwLock has far better perf than the std RwLock.

Seems like the current usage doesn't have any potential poisoning issues though. Maybe the unwraps would be ok.

nrc: I'm no expert on this, but my impression is that you always have to deal with the possibility of inconsistent data caused by a panicking thread and lock poisoning deals with this by panicking other threads that attempt to access that data. If you don't use poisoning, then you need some other strategy or you'll have rare and subtle bugs.

I don't think the performance of r/w lock will be meaningful in Racer, I would expect it to be dwarfed by the other things that happen. But I haven't profiled or anything, and again I'm not an expert.

jwilm: > I don't think the performance of r/w lock will be meaningful in Racer,

You're probably right, and we can always revisit later if needed.

jonasbb:
@jwilm Is there a good way to estimate or measure the overhead? The benchmarks did not compile for me and seem to be outdated.

bkchr:
What is the status on this pull request?

bkchr:
@TedDriggs what do you think of this pr? I could work in the comments and reopen the pull request.

TedDriggs:
If you do, I'd be happy to review it.

Issue #695, opened 2017-02-12
`find-definition` doesn't correctly find closure parameters


mlalic:
Two minimal ways to reproduce the issue.

For:
```rust
fn main() {
    let y = Some(5);
    y.map(|x| x + 1);
}
```

Looking for the definition of `x` in the closure body yields nothing.
```
mlalic@gallifrey:~/src/racer/target/debug$ ./racer find-definition 3 15 test.rs
END
```

For:
```rust
fn main() {
    let x = Some(5);
    x.map(|x| x + 1);
}
```

Again, looking for the definition of `x` *from the closure body* this time yields the `x` in the enclosing scope, which is arguably worse than not finding anything.

```
mlalic@gallifrey:~/src/racer/target/debug$ ./racer find-definition 3 15 test.rs
MATCH x,2,8,test.rs,Let,let x = Some(5);
END
```

TedDriggs:
This appears to be fixed by #697.

PR #697, opened 2017-02-16
Implements the find definition for closures

bkchr:
- Bracket scope and scope without brackets is working
- Adds 12 tests to ensure that everything is working as expected
- Introduces new Closure type for matches

Should fix issue #695.
I will also try to implement to extract type information in another pull-request.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ toml = "0.2.1"
 env_logger = "0.3.4"
 typed-arena = "1.2"
 clap = "2.19"
+regex = "0.2"
 
 [dependencies.clippy]
 version = "0.0.103"
--- a/src/racer/lib.rs
+++ b/src/racer/lib.rs
@@ -11,6 +11,7 @@ extern crate syntex_errors;
 extern crate toml;
 extern crate env_logger;
 extern crate typed_arena;
+extern crate regex;
 
 #[macro_use]
 mod testutils;
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -11,6 +11,7 @@ use cargo;
 use std::path::{Path, PathBuf};
 use std::{self, vec};
 use matchers::PendingImports;
+use regex::Regex;
 
 #[cfg(unix)]
 pub const PATH_SEP: char = ':';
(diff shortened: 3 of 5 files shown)


jwilm:
This looks really good! One concern--Changing `MatchType` is a breaking change. Is there any reason the `Closure` variant is needed versus something like `FnArg`?

bkchr:
No I don't think that it is required at the moment. I thought that it is a cleaner solution for identifying the type. I could change the type to an FnArg. However, for resolving the type of the arguments of the closure, it maybe speed ups the process by directly knowing that we need to parse a closure. 

I think I change the type to FnArg and if the distinction for other feature implementations is required, the new match type could be added. Is that okay?

bkchr:
Okay, I updated the pull request :) More concerns?

bkchr:
Hi, are you still want to merge this?

aeickhoff:
Just wanted to say that in my particular project having support for closure argument type inference would be a huge win! Hoping this PR will go places. I would be completely new to the racer source code, but if you need help @bkchr, let me know!

bkchr:
Yeah this pull request works, the problem is that it only works if you define the argument in the closure definition. So:
|v:i32| works, but |v| does not work. But as far as I understand the rest of racer, that kind of implementation is not that easy.

aeickhoff:
That would already help a lot!

jwilm:
Sorry for the delay on this @bkchr. The PR looks great!

Went to merge this morning; seem to have a conflict with recent changes in Cargo.toml. Not sure why git can't figure this one out on its own...

Anyway, would you mind rebasing your PR? Ready to merge once the button is green :smile:

bkchr:
Done :)

jwilm:
Thanks again; merged!

TedDriggs:
@bkchr, can you explain what that regex is trying to do? I'm investigating #732 and it looks like that is matching things when it shouldn't, but I'm not completely clear on what's happening.

bkchr:
Yeah I think that I found the bug, I missed to exclude ";". I comment in your issue with the possible fix.

TedDriggs:
@jwilm, I've found a place where the distinction does matter. `typeinf::get_type_of_fn_arg` assumes that it's getting a function argument, and sets off looking for the statement start. In #731, that fails.

PR #698, opened 2017-02-17
Fix stack overflows

FraGag:
This is my attempt at fixing the infamous stack overflows (#354). They were crashing the Rust Language Server in my project and I don't like dirty workarounds. 😄

The first commit fixes the legitimate case where an import references a value (function, const, static, etc.) that has the same name as the first component in the import's path. For example:

```rust
use foo::foo;

mod foo {
    pub fn foo() {}
}
```

Racer will now report both the module and the function where applicable.

The second commit fixes unresolvable imports that refer to themselves, as mentioned in https://github.com/phildawes/racer/issues/354#issuecomment-134847072. For example:

```rust
use foo::foo;
// no `foo` defined anywhere
```

To fix this, we now keep track of the imports that are being resolved, and we ignore imports that are being resolved recursively. This has the effect that none of the names in such imports will be reported.

The third commit rolls back a workaround for glob imports causing stack overflows when they were processed recursively. I didn't figure out what caused the condition in the first place, but I suspect that the first two commits would fix it. It also removes the special case of a glob import referencing itself, because the second commit will take care of unbounded recursion.

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -11,6 +11,7 @@ use std::cmp::{min, max};
 use std::iter::{Fuse, Iterator};
 use std::rc::Rc;
 use codeiter::StmtIndicesIter;
+use matchers::PendingImports;
 
 use scopes;
 use nameres;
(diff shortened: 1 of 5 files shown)


FraGag:
cc @jwilm (for #479)

jwilm:
Oh, wow! Thank you very much for putting this together. This ought to make a lot of people (including myself) very happy.

Your implementation looks like exactly the right approach. Going to test this and follow up shortly.

jwilm:
This works really well. Going to merge and make a release immediately. This is **fantastic** work. Thanks again for putting this together!

:star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2: :star2:

PR #701, opened 2017-03-05
Add Kakoune to supported editors

casimir:
Since https://github.com/mawww/kakoune/pull/1249 Kakoune proposes an integration for racer out of the box.

Also I must say Racer is fairly easy to use in CLI, really easy to integrate in editors. Thanks a lot for your work.


jwilm:
Thanks for the PR!

> Also I must say Racer is fairly easy to use in CLI, really easy to integrate in editors. Thanks a lot for your work.

That thanks goes to @phildawes and the other contributors. :smile:

PR #703, opened 2017-03-18
Update atom racer package link

Voultapher:
The current link leads to an old unmaintained and broken atom package.
Luckily there is a good alternative.


jwilm:
Thanks for the patch!

PR #707, opened 2017-04-13
Prevent panic on macros called 'impl*'

jmillikan:
Racer panics if it tries to find impls in a file like [this one in breeze-emu](https://github.com/jonas-schievink/breeze-emu/blob/2c467b8312fa2baf9edc48e66ccbab7f829c3c3b/src/breeze_core/ppu/mod.rs). Changed search_for_impls to exclude macros like search_for_generic_impls.


jwilm:
Thanks for the patch!

PR #712, opened 2017-04-29
Fix nested list markdown

alyssais:
Previously, the a) and b) headers were just interpreted as part of the paragraphs, and wouldn't be indented correctly. This PR fixes it to use proper Markdown nested lists.


jwilm:
Thanks for the patch!

PR #717, opened 2017-05-18
Fix handling of macros named '!'

### LINKED ISSUE #716: Racer is very aggressive about rejecting impls
I was trying to use Racer to complete things in the `gfx` crate. However, it seems that Racer is really conservative about what it parses. For example, it won't provide suggestions for completions for `gfx::Encoder` since part of the body contains a `!`. (see nameres.rs, line 242). It seems like the intention is to only check if `impl` is a keyword, but the current code will reject an `impl` if _anything_ in the body contains `!` which is probably not what we want.

bobtwinkles:
Previously Racer was very aggressive about rejecting things that looked like a macro named `impl!`. However, this would result in Racer rejecting legitimate `impls` like the following:

```rust
impl Inverter {
  pub fn invert(&self, b: bool) -> bool {
    !b
  }
}
```

Fixes #716.

### DIFF
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -240,10 +240,15 @@ pub fn search_for_impls(pos: usize, searchstr: &str, filepath: &Path, local: boo
     for (start, end) in src.iter_stmts() {
         let blob = &src[start..end];
 
-        if blob.starts_with("impl")
-            && !blob.contains('!') {
+        if blob.starts_with("impl") {
             blob.find('{').map(|n| {
-                let mut decl = blob[..n+1].to_owned();
+                let ref decl = blob[..n+1];
+                if decl.contains('!') {
+                    // Guard against macros
+                    debug!("impl was probably a macro: {} {}", filepath.display(), start);
+                    return;
+                }
+                let mut decl = decl.to_owned();
                 decl.push_str("}");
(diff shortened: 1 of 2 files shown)


bobtwinkles:
If desired, I can add a test for this. However, it seemed like a trivial enough change and I banged this out quickly mostly for my own personal use =P.

jwilm:
A test for this behavior would be great! Thanks for the PR as well.

bobtwinkles:
I added two tests, one for the case where we have a valid impl that would previously have been rejected (which fails on current master) and one that tests to make sure we don't reintroduce the that the change that introduced this bug caused.

CryZe:
Okay, this should be merged pretty soon, as racer 2.0.7 regressed immensely and completes almost no impl anymore because of them containing the ! symbol.

cedenday:
How did the regressive pull request get approved anyway?

jwilm:
@bobtwinkles would you mind updating this with the latest master?

@cedenday tests passed. Sometimes bugs get through.

bobtwinkles:
Should be rebased. `tests/system.rs` should probably get split up at some point, git (and I) got totally lost trying to merge it :sweat_smile:.

jwilm:
Thanks for the rebase!

PR #722, opened 2017-05-25
Find types of tuple fields referenced by index

### LINKED ISSUE #721: Racer cannot infer type of fields accessed by index, such as self.0
```rust
pub struct Bar(String);

impl Bar {
    pub fn starts_with(self, needle: &str) -> bool {
        self.0.$
    }
}
```

**Expected:** The cursor, when placed at `$`, should provide type completions for `String`.
**Observed:** None are provided.

This is because `ast::ExprTypeVisitor::visit_expr` does not handle `TupField`.

TedDriggs:
Addresses issue #721 by using logic similar to what was already present for named struct fields


jwilm:
This looks great, thanks!

Issue #719, opened 2017-05-24
Racer should autocomplete methods/fields for struct members that are references


TedDriggs:
On rustc 1.19.0-nightly (f1140a331 2017-05-08) and racer 2.0.7, placing the cursor at the `$` in the following example doesn't provide any method suggestions. This appears to be true regardless of the type the reference identifies.

```rust
pub struct Foo<'a> {
    bar: &'a str,
}

impl<'a> Foo<'a> {
    pub fn starts_with(&self, needle: &str) -> bool {
        self.bar.$
    }
}
```

PR #723, opened 2017-05-25
Completions for reference fields

TedDriggs:
This change will look through references to get the underlying type and provide completions off that, addressing #719.


jwilm:
This looks good. Would you mind adding a test, though? Otherwise this will be susceptible to breakage.

jwilm:
Thanks!

PR #724, opened 2017-05-30
Find field types when bound through let exprs on struct fields

### LINKED ISSUE #720: Racer cannot infer type for fields bound through let patterns
On `rustc 1.19.0-nightly (5b13bff52 2017-05-23)` and racer 2.0.7, the following code doesn't produce expected completions:

```rust
pub struct Foo<'a> {
    bar: &'a str,
    baz: String,
}

impl<'a> Foo<'a> {
    pub fn starts_with(self, needle: &str) -> bool {
        let Foo { bar, baz, .. } = self;

        baz.$
    }
}
```

**Expected:** With the cursor where the `$` is currently, there should be methods for `String`; this can be confirmed by typing `self.baz.` in the same location. However, this pattern binding doesn't produce any completions.

**Observed:** Here's the trace from racer:

```
PREFIX 185,185,
DEBUG:racer::core: Field: contextstr is |baz|, searchstr is ||
DEBUG:racer::ast: visit_expr expr(4294967295: baz)
DEBUG:racer::ast: expr is a path P[baz]
DEBUG:racer::ast: resolve_ast_path P[baz]
DEBUG:racer::nameres: resolve_path_with_str P[baz]
DEBUG:racer::nameres: resolve_path P[baz] "./src/lib.rs" 185 ExactMatch
DEBUG:racer::nameres: resolve_name baz "./src/lib.rs" 185 ExactMatch Both
DEBUG:racer::nameres: search_local_scopes PathSegment { name: "baz", types: [] } "./src/lib.rs" 185 ExactMatch Both
DEBUG:racer::nameres: searching scope Both start: 130 point: 185 'baz' "./src/lib.rs" ExactMatch local: true, session: Session { .. }
DEBUG:racer::matchers: match_pattern_let point is 154
DEBUG:racer::typeinf: get_type_of match Match ["baz", "./src/lib.rs", 154, true, Let, [], [] |let Foo { bar, baz, .. }|]
thread 'searcher' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:329
note: Run with `RUST_BACKTRACE=1` for a backtrace.
ERROR:racer: Search thread paniced: Any
```

Specifically, this line is failing:

```rust
let point = scopes::find_stmt_start(msrc, m.point).unwrap();
```

I'll keep investigating; at minimum, I may submit a PR replacing some `unwrap()` calls with more descriptive `expect` calls.

TedDriggs:
This fixes #720 by improving how Racer identifies the start of `let` expressions. Specifically, it will keep looking for the statement start until it finds a statement that begins `let ...` up to a limit of 6 attempts.

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -255,6 +255,32 @@ fn destructure_pattern_to_ty(pat: &ast::Pat,
                 None
             }
         }
+        PatKind::Struct(ref path, ref children, _) => {
+            let m = resolve_ast_path(path, &scope.filepath, scope.point, session);
+            let contextty = path_to_match(ty.clone(), session);
+            if let Some(m) = m {
+                let mut res = None;
+
+                for child in children {
+                    if point_is_in_span(point as u32, &child.span) {
+                        res = typeinf::get_struct_field_type(&child.node.ident.name.as_str(), &m, session)
+                            .and_then(|ty|
+                                if let Some(Ty::Match(ref contextmatch)) = contextty {
+                                    path_to_match_including_generics(ty, contextmatch, session)
+                                } else {
+                                    path_to_match(ty, session)
+                                })
+                            .and_then(|ty| destructure_pattern_to_ty(&child.node.pat, point, &ty, scope, session));
+
+                        break;
+                    }
+                }
+
+                res
+            } else {
+                None
+            }
+        }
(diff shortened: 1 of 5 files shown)


jwilm:
Looks good; thanks!

Issue #692, opened 2017-01-29
expressions with `?` operator give no suggestions, returning "error: expected expression, found `<eof>`"


imiuka:
heya,

seems that type inference on `?` doesn't work. given the input file:

```rust
use std::error::Error;

fn main() {
    test();
}

fn test() -> Result<&'static str, Box<Error>> {
    let x: Result<&'static str, Box<Error>> = Ok("hello");
    let y = x?.to_uppercase();
    let z = y.

    Ok("")
}
```

racer returns:
```bash
$ racer complete-with-snippet 9 16 main.rs
PREFIX 173,173,
error: expected expression, found `<eof>`

END
```

similarly, completion suggestion for types based on `y` don't work either.

PR #726, opened 2017-06-01
Add Try Operator support

TedDriggs:
This will address #692, but I have some questions about how best to proceed.

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -611,6 +611,12 @@ impl<'c, 's> visit::Visitor for ExprTypeVisitor<'c, 's> {
                 });
             }
 
+            ExprKind::Try(ref expr) => {                
+                debug!("try expr");
+                self.visit_expr(&expr);
+                debug!("{:?}", self.result);
+            }
+
             _ => {
                 debug!("- Could not match expr node type: {:?}",expr.node);
             }
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -624,8 +624,14 @@ impl<'c, 's> visit::Visitor for ExprTypeVisitor<'c, 's> {
                                         &ok_var.filepath, 
                                         ok_var.point, 
                                         self.session)
+                    } else if m.matchstr == "Result" && (m.generic_types.len() != m.generic_args.len()) {
+                        debug!("Unable to desugar Try expression; either `T` or `E` was `()`.");
+                        None
                     } else {
-                        debug!("Unable to desugar Try expression; type was {}", m.matchstr);
+                        debug!("Unable to desugar Try expression; type was {} with arity {} of {}", 
+                            m.matchstr, 
+                            m.generic_types.len(),
+                            m.generic_args.len());
                         None
                     }
                 } else {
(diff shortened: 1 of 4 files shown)


src/racer/ast.rs
```diff
@@ -611,6 +611,12 @@ impl<'c, 's> visit::Visitor for ExprTypeVisitor<'c, 's> {
                 });
             }
 
+            ExprKind::Try(ref expr) => {                
+                debug!("try expr");
+                self.visit_expr(&expr);
+                debug!("{:?}", self.result);
```
TedDriggs: @jwilm, once I know the inner type of `&expr`, how would I convert that to the associated type `Try::Ok`?

src/racer/scopes.rs
```diff
@@ -228,6 +228,7 @@ pub fn get_start_of_search_expr(src: &str, point: usize) -> usize {
             (b'.', State::MustEndsWithDot(_)) =>  State::None,
             (b':', State::MustEndsWithDot(index)) =>  State::StartsWithCol(index),
             (b':', State::StartsWithCol(_)) =>  State::None,
+            (b'?', State::StartsWithDot) => State::None,
```
TedDriggs: As written, this will only work for method or field accesses on a `Try` operator. Should `?` be in [`util::is_search_expr_char`](https://github.com/phildawes/racer/blob/master/src/racer/util.rs#L14) instead?

TedDriggs:
This is passing on stable and nightly on my box right now; not sure why the tests are failing on Travis yet.

src/racer/scopes.rs
jwilm: I think the current impl is fine. `?` isn't really a search expr char, there's no completions to make until you attempt field access.

tests/system.rs
```diff
@@ -2908,6 +2908,61 @@ fn ignores_impl_macro() {
...
+            }
+        }
+
+        pub fn half(val: Foo) -> Result<Foo, OddError> {
+            Ok(Foo(be_even(val)?.~0 / 2))
+        }
+    ";
+
+    let got = get_definition(src, None);
+    assert_eq!("0", got.matchstr);
```
jwilm: What do you think about adding a completions test with a non-tuple struct? I imagine like it works, but these tests don't cover that case.

TedDriggs: Will do.

src/racer/ast.rs
```diff
@@ -611,6 +611,28 @@ impl<'c, 's> visit::Visitor for ExprTypeVisitor<'c, 's> {
...
             }
 
+            ExprKind::Try(ref expr) => {                
+                debug!("try expr");
+                self.visit_expr(&expr);
+                self.result = if let Some(&Ty::Match(ref m)) = self.result.as_ref() {
+                    // HACK: Try to break open the result and find it's "Ok" type.
+                    // Once the 'Try' operator trait stabilizes, it'd be better to
+                    // find the type through the trait.
+                    if m.matchstr == "Result" && m.generic_types.len() == 2 {
```
TedDriggs: @jwilm I found a problem with this line; `()` doesn't appear in the `generic_types` list, so `Result<_, ()>` has `generic_types.len() == 1`. Even worse, the same is true of `Result<(), _>`; in that case, `generic_types[0]` gets the `E` case, rather than the `T` case. Is this deliberate, or a bug somewhere else in racer?

Issue #443, opened 2015-12-10
String literal completions


kybernetikos:
As a newcomer to rust, I benefit massively from racer helping me to discover the sorts of things I can do.

This makes me sad:

![nostringcompletions](https://cloud.githubusercontent.com/assets/430980/11731822/3e2c8b38-9f96-11e5-88d9-ad00535678fa.png)

I get completions for other things, like the vector example in the main readme, so I don't think it's just badly set up.

I'm using a cargo installed version of racer, with a nightly rust build and my rust src is set.  The screenshot is from Eclipse DT.

phildawes:
Hi @kybernetikos. Thanks for your issue, this is in the pipeline. I'm working on getting racer to use rustc to evaluate the types of expressions, which will include literals. It's a bit hard going because the compiler isn't designed to be used in this way, but it will allow completions for any legal expression once done.

kybernetikos:
that's good news.  thanks for all your work.

TedDriggs:
The fix for this is now in master, and should be included in the next release.

PR #728, opened 2017-06-05
Add method autocompletions on common string literals

TedDriggs:
This adds partial support for method completions on string literals (requested in #443), with some caveats:

1. Raw strings aren't supported
2. Escaped quotation marks in cooked strings could cause problems

Both of these are omitted because they would require looking ahead one or more characters, which the loop in `scopes:: get_start_of_search_expr` isn't set up for right now.

### DIFF
--- a/tests/system.rs
+++ b/tests/system.rs
@@ -3046,3 +3046,16 @@ fn closure_bracket_scope_nested_match_outside() {
     assert_eq!("x", got.matchstr);
     assert_eq!("| x: i32 |", got.contextstr);
 }
+
+#[test]
+fn literal_string_method() {
+    let _lock = sync!();
+    let src = r#"
+        fn check() {
+            "hello".st~arts_with("he");
+        }
+    "#;
+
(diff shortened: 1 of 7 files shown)


TedDriggs:
@jwilm I fixed the build break, but for some reason it's not running the CI build this time.

jwilm:
I poked travis

tests/system.rs
```diff
@@ -3045,4 +3045,17 @@ fn closure_bracket_scope_nested_match_outside() {
...
+fn literal_string_method() {
+    let _lock = sync!();
+    let src = r#"
+        fn check() {
+            "hello".st~arts_with("he");
+        }
+    "#;
+
+    let got = get_definition(src, None);
+    assert_eq!("starts_with", got.matchstr);
```
jwilm: May also want to add a completions test here.

TedDriggs: I'll add some more tests and then merge the change.

PR #734, opened 2017-06-08
Be stricter about deciding what constitutes a closure

### LINKED ISSUE #732: Racer 2.0.8 finds incorrect definition match
For this a.rs:
```rust
fn foo() {
    i
}
fn bar() {
    let i = 0;
    let x = 0 | i;
}
fn baz() {
    // 1 || 2;
}
```

I get the following output:
```
$ racer find-definition 2 5 a.rs
MATCH i,6,16,a.rs,FnArg,| i; } fn baz() { // 1 |
END
```
This can make racer unusable in emacs as these definitions can be very long and match very common patterns; in this example, every time a symbol starts with `i`, the definition appears when `i` is typed in.

TedDriggs:
This fixes #732 by separating the "look for a closure declaration" from "look for ident inside closure parameters". It's probably still not optimal, but it passes the new tests and seems accurate in my local usage.

@bkchr, please have a look at the regexes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ doc = false
 debug = true
 
 [dependencies]
+lazy_static = "0.2"
 log = "0.3.6"
 syntex_syntax = "0.52.0"
 syntex_errors = "0.52.0"
@@ -36,7 +37,6 @@ optional = true
 
 [dev-dependencies]
 rand = "0.3"
-lazy_static = "0.2"
 
 [features]
 nightly = []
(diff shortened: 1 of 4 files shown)


bkchr:
Yeah I think your new regex is much better and simpler. If we just use your regex, we could skip my regex in the second script and just search for the variable definition. Because we already know that we are in a closure definition, we could skip all the rest.

TedDriggs:
@bkchr, when I tried to simplify your regex, I started failing the test I added `closure_scope_dont_match_type_annotations`; that's why I left it as-is. I'm submitting a new PR now that puts the first regex into `lazy_static`.

bkchr:
Couldn't you use the start and end of the ||, found with your regex, to create a substring. Use that substring with the txt_matches function. Shouldn't that work?

PR #735, opened 2017-06-09
Replace `unwrap` with `expect` in racer::typeinf

TedDriggs:
Replaced a bunch of `unwrap` calls with more descriptive `expect` statements.

Also added a new public function `scopes::expect_stmt_start` to avoid repeating the same error in lots of places.


jwilm:
looks good to me!

PR #737, opened 2017-06-09
Search for functions on enums in addition to finding variants

TedDriggs:
Previously, racer would _only_ return enum variants. However, enums are also allowed to have `impl` blocks and can declare static functions. This change uses the struct search code on enums, and will always return variants followed by impl functions.


jwilm:
Looks good to me, although I think we should add a test for variant/impl ordering in completions.

PR #740, opened 2017-06-12
Find doc strings for `let` statements

TedDriggs:
This adds doc finding for values defined through local `let` statements.

Flagging for @imperio, @misdreavus, @Ralith for thoughts, as it's not clear how idiomatic this behavior is in Rust today.


TedDriggs:
This is now very out-of-date; going to close it and will reopen in future.

PR #745, opened 2017-06-15
Update README.md

Chillee:
The old author has disappeared, and this is the currently maintained extension for VSCode.


jaxx:
Same as https://github.com/racer-rust/racer/pull/738 in my opinion.

Chillee:
Ah yes that's true. Thanks!

TedDriggs:
@Chillee thanks! Please have a look at #738 and let me know if you have any concerns with the proposed wording there.

Issue #165, opened 2015-02-25
Hide inaccessible members and functions


Stebalien:
Currently, racer completes private struct members and functions cross module.

tafia:
@phildawes just to be sure, everything should be done in matchers.rs ?
With `local` variable and `pub` keyword ?

phildawes:
Hi @tafia,
I think for field the problem is in nameres::search_struct_fields() and ast::parse_struct_fields(). It should be using the 'visibility' attribute in the parsed ast 'NamedField' type.

llogiq:
But this would require getting the Visibility from the Item in case the field visibility is inherited. How would one do that?

Phrohdoh:
Has this been deemed a wontfix or is it just not of high importance?

(racer 1.2.10 fwiw)

0xd4d:
Would be nice if this 5 year old issue got fixed since internal functions shouldn't be shown when referencing some other crate.

PR #749, opened 2017-06-19
Support completions for pub(restricted) items

### LINKED ISSUE #748: Racer should find restricted-visibility members
On racer 2.0.8, the following code won't produce a completion:

```rust
mod foo { 
    pub(crate) fn do_stuff() -> String {
        println!("Hello");
        "Hello".into()
    }

    pub fn more_stuff() {
        do_~
    }
}
```

Removing the `(crate)` modifier from the function signature causes the completion to appear as expected.

This issue does **not** cover hiding members when searching for completions outside their scope; that is #165. An acceptable fix for this issue can ignore visibility restrictions.

TedDriggs:
Fixes #748 by adding `pub(restricted)` support to:

* Functions
* Struct fields
* Static methods
* Self-accepting methods

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Support completion after using try operator `?` #726
 - Find methods on cooked string literals #728
 - Find doc comments on named and indexed struct fields #739
+- Find `pub(restricted)` items #748
 
 ## 2.0.8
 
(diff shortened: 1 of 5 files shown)


jwilm:
Looks good at first glance! It appears to be missing negative completion tests; i.e. `pub(in foo) item` shouldn't complete in `mod bar`. Should we add that now or in a separate PR?

TedDriggs:
@jwilm - I specifically didn't try to support comprehension of the visibility restrictions in this PR; that feels better suited for #165.

jwilm:
Ah, right. Sounds like a plan!

Issue #754, opened 2017-07-08
match prevents racer returning correct positions for enums


Wilfred:
Given the code:

```rust
enum Fruit {
    Apple = 1,
}

fn foo(ty: Fruit) -> bool {
    (1 as u8 | Fruit::Apple as u8) == Fruit::Apple as u8
}

fn bar(ty: Fruit) -> bool {
    match ty {
        Fruit::Apple |
        Fruit::Apple => {
            false
        }
    }
}
```

Racer cannot find the definition of `Fruit` when called from the body of `foo`:

```
$ RUST_SRC_PATH=/home/wilfred/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src racer find-definition 6 20 foo.rs 
MATCH Fruit,6,1,foo.rs,FnArg,| Fruit::Apple as u8) == Fruit::Apple as u8 } fn bar(ty: Fruit) -> bool { match ty { Fruit::Apple |
END
```

The position returned by racer is whitespace on the wrong line.

TedDriggs:
This seems to be another flavor of the issue #734 tried to address; it's mistaking the `|` characters for a closure. Here's the log line which shows the offending "match"

```
TRACE:racer::nameres: Found a closure declaration 72..184, about to search for
 `Fruit` in | Fruit::Apple as u8) == Fruit::Apple as u8
}

fn bar(ty: Fruit) -> bool {
    match ty {
        Fruit::Apple |
```

Wilfred:
The really weird thing about this bug is that it occurs even when the `bar` function is commented out:

```rust
enum Fruit {
    Apple = 1,
}

fn foo(ty: Fruit) -> bool {
    (1 as u8 | Fruit::Apple as u8) == Fruit::Apple as u8
}

// fn bar(ty: Fruit) -> bool {
//     match ty {
//         Fruit::Apple |
//         Fruit::Apple => {
//             false
//         }
//     }
// }
```

However, removing the `bar` function prevents the bug.

TedDriggs:
I think that's still a result of the closure issue. @jwilm I don't have time to do a rewrite of the closure work that's more resilient at the moment. It seems to work okay some of the time, but there are definitely edge cases like this that keep popping up. Would you rather we keep applying stopgap fixes for now, or rip it out completely?

jwilm:
[...]

PR #756, opened 2017-07-13
Rewrites closure detection to work without using regular expressions

bkchr:
- Remove all regex occurrences
- Use the syntex crate to parse the actual closure parameters
- Add test for issue 754

Fixes: #754

### DIFF
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -5,13 +5,12 @@ use core::SearchType::{self, ExactMatch, StartsWith};
 use core::{Match, Src, Session, Coordinate, SessionExt, Ty, Point};
 use core::MatchType::{Module, Function, Struct, Enum, FnArg, Trait, StructField, Impl, TraitImpl, MatchArm, Builtin};
 use core::Namespace;
-use util::{symbol_matches, txt_matches, find_ident_end};
+use util::{closure_valid_arg_scope, symbol_matches, txt_matches, find_ident_end};
 use matchers::find_doc;
 use cargo;
 use std::path::{Path, PathBuf};
 use std::{self, vec};
 use matchers::PendingImports;
-use regex::Regex;
 
 #[cfg(unix)]
 pub const PATH_SEP: char = ':';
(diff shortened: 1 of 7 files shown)


src/racer/util.rs
```diff
@@ -70,6 +70,27 @@ pub fn symbol_matches(stype: SearchType, searchstr: &str, candidate: &str) -> bo
     }
 }
 
+// Try to valid if the given scope contains a valid closure arg scope.
```
TedDriggs: NIT: Please use `///` so this gets picked up as a doc comment.

src/racer/util.rs
```diff
@@ -70,6 +70,27 @@ pub fn symbol_matches(stype: SearchType, searchstr: &str, candidate: &str) -> bo
...
+    let right_pipe = if let Some(pos) = rest_scope.find('|') {
+        left_pipe + 1 + pos
+    } else {
+        return None;
+    };
+
+    let pipe_scope = &scope_src[left_pipe..right_pipe+1];
+
+    // If we found not forbidden characters --> no closure definition
+    let forbidden_chars = [ ";", "{", "}" ];
```
TedDriggs: Struct patterns are allowed in closure arguments; **[here's](https://play.rust-lang.org/?gist=c5ad305c6c473028507a1926ff225664&version=stable)** an example. Please either a) have that work, and add a test to demonstrate it, or b) add a note to this method's doc comment calling out that struct patterns in closure arguments aren't supported.

tests/system.rs
```diff
@@ -3330,6 +3330,34 @@ fn closure_bracket_scope_nested_match_outside() {
 }
 
 #[test]
+fn issue_754() {
```
TedDriggs: Please make the test name descriptive of what it demonstrates. Adding a link to the GitHub issue in the test's doc comment would be great though 😃

src/racer/nameres.rs
```diff
@@ -946,68 +946,49 @@ pub fn search_scope(start: Point, point: Point, src: Src,
...
-                local: true,
-                mtype: FnArg,
-                contextstr: cap.get(0).unwrap().as_str().to_owned(),
-                generic_args: Vec::new(),
-                generic_types: Vec::new(),
-                docs: String::new(),
-            })
-        } else {
-            None
+    if let Some((left_pipe, _, pipe_scope)) = closure_valid_arg_scope(scope_src) {
```
TedDriggs: Can you please add a `trace!` statement here which reports that the function has found a closure?

bkchr: Added :)

bkchr:
Addressed your requests. Especially thanks for finding that curly bracket use case in a closure argument scope :D

TedDriggs:
Thanks for adding the tests! I'd like @jwilm to look at it, but otherwise LGTM

bkchr:
Ping @jwilm ^^

bkchr:
Fixed the merge conflict! So it should be ready to merge! :)

PR #764, opened 2017-07-17
Fix #633 by looking for statement starts in path resolution

### LINKED ISSUE #633: Does not complete after newline inside `use module::{`
I really enjoy racer nowadays, it works very well.

Here's a case for `use` imports that it does not complete correctly:

``` rust
use visit::{GraphBase, IntoNeighbors,                                            
    In                                                                           
      ^    
```

With the cursor at the ^ location, it doesn't complete from the `visit` module anymore, even though it's still inside the same `{..}` bracket from use.

TedDriggs:
This looks to address #633. That issue was the result of looking back to the start of the _line_ when performing path completions. The code was apparently under the impression that a line break could not occur inside a path. While this may be true in normal paths, newlines can occur in `use` statements.

The fix looks for the statement start when performing a path completion and then falls back to looking for the line start if it can't find a statement.

**Caution:** This seems to pass all the tests, but we don't have performance benchmarks and this change is likely to be hit in a _lot_ of cases.

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -915,7 +915,7 @@ fn complete_from_file_(
     session: &Session
 ) -> vec::IntoIter<Match> {
     let src = session.load_file_and_mask_comments(filepath);
-    let src = &src.as_src()[..];
+    let src_text = &src.as_src()[..];
 
     // TODO return result
     let pos = match cursor.to_point(&session.load_file(filepath)) {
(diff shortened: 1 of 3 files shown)


jwilm:
This looks good to me, but there's no way to comment on perf without at least a simple benchmark.

TedDriggs:
I'm not familiar with `cargo bench`, do you want me to hold off on the commit until we have benchmarks, or go for correctness and if it causes slowdowns then we can revisit?

jwilm:
I'd lean towards merging it and seeing if it causes any issues. We can always optimize later if needed.

TedDriggs:
If you mark your review as "approved" then I'll go ahead and do the merge.

jwilm:
done

PR #767, opened 2017-07-19
Honor `use as` aliases when suggesting completions

### LINKED ISSUE #753: Racer should recognize use alias in braces
My racer version is 2.0.9. Here's an example.
```rust
mod m {
    pub struct Wrapper {
        pub x: i32,
    }
}

fn main() {
    use m::{Wrapper as Wpr}; // here
    let s = Wpr{ x: 1 };
    W~; // at `~`, always suggest `Wrapper` rather than `Wpr`
    s.~ // at `~`, no autocomplete unless remove the braces above
}
```

### LINKED ISSUE #758: When importing via "use as", completion should offer the "as ___" name
On racer v2.0.9:

```rust
mod m {
    pub struct Second {
        pub y: i32,
    }
}

fn foo() {
    use m::Second as Repro;
    Rep~
}
```

**Expected:** There should be a match for `Repro` at the `~` character in the input.
**Observed:** No match.

Possibly related to #753

TedDriggs:
This fixes #758 by remembering the alias and then swapping out `matchstr` in the found match with the in-scope identifier.

This change did require altering two tests, as the previous behavior here was to return the actual name, rather than the alias.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 - Support resolving `use as` aliases declared in multi-element `use` statements #753
 - Provide suggestions for global paths in more cases #765
+- Suggestions imported via `use as` statements now return their in-scope alias as the match string #767
 
 ## 2.0.9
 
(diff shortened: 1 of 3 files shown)


CryZe:
Wipe your build caches in order for rustup to update. The `.cargo` folder contains binaries like rustup, but is also part of the build caches on Travis CI, which messes with rustup's update process. So you are still on 1.4.0, which can't handle the beta toolchain anymore.

TedDriggs:
Thanks @CryZe! I've just tried that and restarted the build.

Issue #771, opened 2017-07-22
Add completion for struct impls in submodules


ReeceStevens:
Racer does not offer completion for struct impls if they are defined in submodules. This presents problems when working with generated libraries such as those built by [`svd2rust`.](https://github.com/japaric/svd2rust)

Relevant test case:
```rust
#[test]
fn completes_super_impl_fn() {
    let _lock = sync!();

    let src = r#"
    pub struct A { }

    pub mod a {
        impl super::A {
            pub fn return_number(&self) -> u8 {
                42
            }
        }

    }

    fn main() {
        let some_struct = A { };
        some_struct.~return_number()
    }
    "#;

    within_test_project(|| {
        let all = get_all_completions(src, None);
        println!("{:?}", all);
        let got = get_one_completion(src, None);
        assert_eq!("return_number", got.matchstr);
    })
}
```
I'm working on developing a solution for this, but I'm a fairly new Rustacean and could use a bit of direction if somebody has an idea of how to implement it. Feedback would be appreciated!

TedDriggs:
I strongly suspect racer never even looks in submodules, since doing so could delay our time-to-results. If you run `RUST_LOG=racer=trace racer complete ...` you should be able to watch it travel through your code as it locates the struct, searches for impls in the current module, then returns.

This isn't a pattern I've encountered before in the wild. Could you post a gist or something showing the code generated by `svd2rust`?

ReeceStevens:
Hey @TedDriggs, thanks for the speedy response (sorry I didn't reciprocate that favor!). The debug log for the completion:
```
[reece@archtop src]$ RUST_LOG=racer=trace racer complete 17 24 main.rs
PREFIX 243,243,
DEBUG:racer::core: Field: contextstr is |some_struct|, searchstr is ||
DEBUG:racer::ast: visit_expr expr(4294967295: some_struct)
DEBUG:racer::ast: expr is a path P[some_struct]
DEBUG:racer::ast: resolve_ast_path P[some_struct]
DEBUG:racer::nameres: resolve_path_with_str P[some_struct]
DEBUG:racer::nameres: resolve_path P[some_struct] "main.rs" 243 ExactMatch
DEBUG:racer::nameres: resolve_name some_struct "main.rs" 243 ExactMatch Both
DEBUG:racer::nameres: search_local_scopes PathSegment { name: "some_struct", types: [] } "main.rs" 243 ExactMatch Both
[...]

PR #772, opened 2017-07-22
Implement completion for struct impls defined in submodules

ReeceStevens:
See issue #771

### DIFF
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -257,6 +257,8 @@ pub fn search_for_impls(pos: Point, searchstr: &str, filepath: &Path, local: boo
                     let is_trait_impl = implres.trait_path.is_some();
                     let mtype = if is_trait_impl { TraitImpl } else { Impl };
 
+                    // TODO: figure out why super::attrs aren't getting
+                    // parsed or found as matches here.
                     implres.name_path.map(|name_path| {
                         name_path.segments.last().map(|name| {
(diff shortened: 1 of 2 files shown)


ReeceStevens:
I'm still working on a solution for this, but I wanted to get a PR up early in case anybody had comments or suggestions on how to proceed with this ticket.

bkchr:
Hi, I just looked at the implementation and did not test anything, but I assume that the following call https://github.com/ReeceStevens/racer/blob/c615f7766ec34b9fce4a3a17af5bc4f769e83961/src/racer/nameres.rs#L254 does not return true. I would start to look there :)

bkchr:
Forget what I said... :( 
Maybe you should look at the blobs and see, in the first place, if the code finds the impl block at all.

ReeceStevens:
@bkchr thanks for the response! @TedDriggs suggested I look through the debug logs and start there, so that's where the investigation shall begin. I'll probably keep discussion on the issue until there's some semblance of a reviewable PR here :)

jwilm:
Removed the `needs review` label until it's not `not ready`. :wink:

TedDriggs:
Last I heard, @ReeceStevens was pursuing an alternate solution to his particular problem with the specific library he's trying to use.

ReeceStevens:
Hey @TedDriggs sorry for the delay in getting back to you. The alternate solution you suggested worked well and I will be opening a PR on `svd2rust` soon. Thanks for your help :)

PR #773, opened 2017-07-22
Fix location where rustup will install sources

wkunkel:
The README currently claims that `rustup component add rust-src` will install sources in `~/.multirust/toolchains/[your-toolchain]/lib/rustlib/src/rust/src`. This is no longer true: sources are installed in `~/.rustup/toolchains/[your-toolchain]/lib/rustlib/src/rust/src`. It seems more robust to use `rustc --print sysroot`, however.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,13 +30,13 @@ As mentioned in the command output, don't forget to add the installation directo
 
 1. Fetch the Rust sourcecode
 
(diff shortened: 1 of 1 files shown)


jwilm:
This seems way better than having a hardcoded path in the README. Thanks!

PR #774, opened 2017-07-24
Fix mod suppression when same-name trait method exists

### LINKED ISSUE #680: Module autocompletions are now incorrectly NOT shown when same-name function exists
This appears to be new to 2.0.4

```rust
use std::fmt;

/// Valid orders during build seasons.
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub enum BuildCommand {
    Build,
    Disband,
}

impl fmt::Display for BuildCommand {
    fn fmt(&self, f: &mut fmt::)
}
```

screen shot 2017-01-05 at 1 45 02 pm

**Expected:** The types in `std::fmt::` should be shown in the suggestions list.

**Note:** Here's what I'm seeing before `fmt::` is typed in:

screen shot 2017-01-05 at 1 48 47 pm

TedDriggs:
Fix #680: Only show trait method suggestions when contextually relevant by checking if the cursor is in a function identifier, and otherwise not looking for trait method declarations and trait methods when searching for unprefixed values.

This is most useful when dealing with `std::fmt`; the example code for this module typically imports it at the mod level, which causes autocomplete not to work for writing the rest of the signature, where `fmt::Formatter` and `fmt::Result` will appear.

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -97,6 +97,19 @@ pub struct Match {
     pub docs: String,
 }
 
+impl Match {
+    /// Checks if two matches can be considered the same for deduplication purposes.
+    ///
+    /// This could be the basis for a `PartialEq` implementation in the future,
+    /// but in the interest of minimizing the crate's public API surface it's exposed
+    /// as a private method for now.
+    fn is_same_as(&self, other: &Match) -> bool {
+        self.point == other.point 
+        && self.matchstr == other.matchstr
+        && self.filepath == other.filepath
+    }
+}
+
 /// The cursor position used by public search methods
 #[derive(Debug, Clone, Copy)]
 pub enum Location {
@@ -819,9 +832,11 @@ pub fn complete_fully_qualified_name<'c, S, P>(
     where S: AsRef<str>,
           P: AsRef<path::Path>,
 {
-    let matches = complete_fully_qualified_name_(query.as_ref(), path.as_ref(), session);
+    let mut matches = complete_fully_qualified_name_(query.as_ref(), path.as_ref(), session);
+    matches.dedup_by(|a, b| a.is_same_as(b));
+    
     MatchIter {
-        matches: matches,
+        matches: matches.into_iter(),
         session: session
     }
 }
@@ -831,7 +846,7 @@ fn complete_fully_qualified_name_(
     query: &str,
     path: &path::Path,
     session: &Session
-) -> vec::IntoIter<Match> {
+) -> Vec<Match> {
     let p: Vec<&str> = query.split("::").collect();
 
     let mut matches = Vec::new();
@@ -855,7 +870,7 @@ fn complete_fully_qualified_name_(
         }
     }
 
-    matches.into_iter()
+    matches
 }
 
 
@@ -902,10 +917,12 @@ pub fn complete_from_file<'c, P, C>(
     where P: AsRef<path::Path>,
           C: Into<Location>
 {
-    let matches = complete_from_file_(filepath.as_ref(), cursor.into(), session);
+    let mut matches = complete_from_file_(filepath.as_ref(), cursor.into(), session);
+    matches.dedup_by(|a, b| a.is_same_as(b));
+
     MatchIter {
+        matches: matches.into_iter(),
         session: session,
-        matches: matches,
     }
 }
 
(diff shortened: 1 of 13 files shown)


TedDriggs:
@jwilm I merged when I should have rebased, sorry for the noise in the PR.

src/racer/nameres.rs
```diff
@@ -1418,6 +1417,68 @@ pub fn resolve_path(path: &core::Path, filepath: &Path, pos: Point,
     }
 }
 
+pub(crate) fn resolve_method(point: Point, msrc: Src, searchstr: &str,
```
jwilm: Would love to have a doc comment on this explaining what the algorithm does.

jwilm: Oh, I see that part of it was just a refactor from elsewhere...

src/racer/matchers.rs
```diff
@@ -23,7 +23,7 @@ pub type PendingImports<'stack, 'fp> = StackLinkedListNode<'stack, PendingImport
 pub type MIter = option::IntoIter<Match>;
 pub type MChain<T> = iter::Chain<T, MIter>;
 
-// Should I return a boxed trait object to make this signature nicer?
+// TODO change return type to `impl Iterator<Item = Match>`
```
jwilm: So excited for this.

src/racer/core.rs
```diff
@@ -955,6 +956,25 @@ fn complete_from_file_(
             let is_global = expr.starts_with("::");
             let is_use = line.starts_with("use ");
 
+            // when in the function ident position, only look for methods
+            // from a trait to complete.
+            if util::in_fn_name(line) {
+                trace!("Path is in fn declaration: `{}`", expr);
+
+                for m in nameres::resolve_method(
```
jwilm: The formatting on this is really hard to parse. It might be better to assign the returned Iter/IntoIter to a temporary variable first. Eg

```rust
let methods = nameres::resolve(/* ... */);
for m in methods {
    out.push(m);
}
```

Actually, maybe it would be better if resolve_method just returned a `Vec` instead of an iterator so that it may be appended directly only `out` instead of doing this iterative thing.

src/racer/nameres.rs
```diff
@@ -525,6 +495,34 @@ fn search_scope_headers(point: Point, scopestart: Point, msrc: Src, searchstr: &
     Vec::new().into_iter()
 }
 
+/// Checks if a scope preblock is a function declaration.
+///
+/// TODO: Handle `extern` functions
+fn preblock_is_fn(preblock: &str) -> bool {
+    // Perform simple checks
+    if preblock.starts_with("fn") || preblock.starts_with("pub fn") || preblock.starts_with("const fn") {
```
jwilm: Iterative approach might be cleaner here. _llvm_ should unroll the loop anyhow. Eg

```rust
for case in &["fn", "pub fn", "const fn"] {
    if preblock.starts_with(case) {
        return true;
    }
}
```

tests/system.rs
```diff
@@ -1285,29 +1285,37 @@ fn keeps_newlines_in_external_mod_doc() {
     assert_eq!("The mods multiline documentation\n\nwith an empty line", got.docs);
 }
 
+/// Addresses https://github.com/racer-rust/racer/issues/618
 #[test]
-fn issue_618() {
+fn always_get_all_doc_lines() {
```
jwilm: It's kind of nice to have the issue number in the test name. I like what you've done to explain what the test actually is though. Maybe we can combine them? `issue_618_always_foo`.

TedDriggs: I moved the issue number to the doc comment so that I could include the full URL. I'm okay with putting issue numbers in test names for the purposes of `cargo test 618` but without easy reference to the issue I'm not sure how valuable the number is.

src/racer/nameres.rs
TedDriggs: True; I wasn't sure at first if they were all going to be the same. I'll change this.

src/racer/core.rs
TedDriggs: As part of #778 I changed the return type of this function to be `Vec`, which will remove the `into_iter` thing. There's honestly no need to append in this case, so I can remove the loop outright and just return the `resolve_method` return value directly.

tests/system.rs
jwilm: For me, it's more about knowing it's associated with 618 when it fails just by looking at cargo test output.

src/racer/nameres.rs
TedDriggs: I added a comment explaining it a bit better.

TedDriggs:
@jwilm I've added the tests against `std::fmt` and addressed all points on the CR except for issue IDs in test names. If it's okay with you, I'll move forward as-is and then we can put in a separate PR to add those back to the test names.

TedDriggs:
Git has gone beserk; closing this PR and making a new one.

PR #778, opened 2017-07-26
Dedup matches in complete

### LINKED ISSUE #645: Duplicate results from racer
Given the code:

```rust
pub enum Level {
}
```

Racer gives:

```
$ CARGO_HOME=/home/wilfred/.cargo RUST_SRC_PATH=/home/wilfred/src/rustc-1.10.0/src /home/wilfred/.cargo/bin/racer complete-with-snippet 1 14 /home/wilfred/projects/bfc/src/diagnostics.rs
PREFIX 9,14,Level
MATCH Level;Level;1;9;/home/wilfred/projects/bfc/src/diagnostics.rs;Enum;pub enum Level {;""
MATCH Level;Level;1;9;/home/wilfred/projects/bfc/src/diagnostics.rs;Enum;pub enum Level {;""
END
```

TedDriggs:
Fix #645 by deduping adjacent matches in the `complete` functions. Matches are deduplicated by filepath, point in file, and matchstr. This tuple should be sufficient to uniquely identify a match.

There is still the possibility of duplicate matches if the duplicates are not adjacent in the returned results. That condition did not arise during testing.

This doesn't try to stop racer from finding items more than once; it instead just removes them before returning. This approach seemed to be the less invasive and lower-risk fix.

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -97,6 +97,15 @@ pub struct Match {
     pub docs: String,
 }
 
+impl Match {
+    /// Checks if two matches can be considered the same for deduplication purposes.
+    fn is_same_as(&self, other: &Match) -> bool {
+        self.point == other.point 
+        && self.matchstr == other.matchstr
+        && self.filepath == other.filepath
+    }
+}
+
 /// The cursor position used by public search methods
 #[derive(Debug, Clone, Copy)]
 pub enum Location {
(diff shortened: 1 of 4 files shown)


src/racer/core.rs
```diff
@@ -97,6 +97,15 @@ pub struct Match {
     pub docs: String,
 }
 
+impl Match {
+    /// Checks if two matches can be considered the same for deduplication purposes.
+    fn is_same_as(&self, other: &Match) -> bool {
```
jwilm: Is there a reason we don't just use PartialEq/Eq here? From the perspective of the user, they are identical.

TedDriggs: I didn't want to publicly expose the trait. If we ever had to remove it in the future, it'd be a breaking change.

jwilm: Makes sense to me. Glad you're thinking about stable API. :smile:

PR #784, opened 2017-08-07
Provide completions for associated types

### LINKED ISSUE #775: Provide completions for associated types in trait impls
```
pub struct Foo;

impl ::std::str::FromStr for Foo {
    type ~Err = ();
    // ... remainder elided
}
```

**Desired Behavior:** When the cursor is at `~` in the example above, calling `racer complete` should list the associated types for the enclosing trait.

TedDriggs:
Fixes #775. This fix is very similar to the one for trait functions:

1. Look to see if cursor is after a declaration keyword
2. If so, try to find the enclosing trait
3. If trait found, look for trait items of the right type

### DIFF
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -3,7 +3,7 @@
 use {core, ast, matchers, scopes, typeinf};
 use core::SearchType::{self, ExactMatch, StartsWith};
 use core::{Match, Src, Session, Coordinate, SessionExt, Ty, Point};
-use core::MatchType::{Module, Function, Struct, Enum, FnArg, Trait, StructField, Impl, TraitImpl, MatchArm, Builtin};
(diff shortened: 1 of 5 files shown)


src/racer/nameres.rs
```diff
@@ -233,6 +233,43 @@ fn search_scope_for_static_trait_fns(point: Point, src: Src, searchstr: &str, fi
...
+    let mut out = Vec::new();
+    for (blobstart,blobend) in scopesrc.iter_stmts() {
+        let blob = &scopesrc[blobstart..blobend];
+        blob.find(|c| c == '=' || c == ';').map(|n| {
+            let signature = blob[..n].trim_right();
+
+            if txt_matches(search_type, &format!("type {}", searchstr), signature) {
+                debug!("found associated type starting |{}| |{}|", searchstr, blob);
+                // TODO: parse this properly
+                let start = blob.find(&format!("type {}", searchstr)).unwrap() + 5;
```
jwilm: Since `format!` allocates, is kind of expensive, and this same string is generated in two places, it should probably be made a variable binding above.

PR #792, opened 2017-08-30
Added workspace support.

### LINKED ISSUE #687: can not work with workspace
Step to reproduce:

1. Create directory `workspace` and in this directory create simple project with `cargo new --bin rusqlite_not_found`,
add  line `extern crate rusqlite;` to src/main.rs and line `rusqlite = {version = "0.8.0", features = ["bundled"]}`
to Cargo.toml (to `dependicies` section).
2. Try `find definition` functionality of racer on word `rusqlite` in line `extern crate rusqlite;` in file `main.rs`.  
**Functionlity works as expected**
3. Remove `target` and `Cargo.lock` files, and add Cargo.toml with such content:
```
[workspace]
members = ["rusqlite_not_found"]
``` 
into `workspace` directory (on one level above of `rusqlite_not_found`)

at now step 2 not work for me (racer 2.0.4).

GoldenBadger:
When searching for a lockfile, get_crate_file now searches upwards through the directory tree looking for a Cargo.toml file containing a `[workspace]` section. If it finds one, it then locates the corresponding Cargo.lock file.

This resolves #687 and is a replacement PR for #649.

### DIFF
--- a/src/racer/cargo.rs
+++ b/src/racer/cargo.rs
@@ -705,8 +705,8 @@ pub fn get_crate_file(kratename: &str, from_path: &Path) -> Option<PathBuf> {
 
         let mut lockfile = tomlfile.clone();
         lockfile.pop();
-        if workspace.is_some() {
-            lockfile.push(workspace.unwrap());
+        if let Some(workspace) = workspace {
+            lockfile = workspace;
             trace!("get_crate_file workspace is {:?}", lockfile);
         }
         lockfile.push("Cargo.lock");
--- a/src/racer/cargo.rs
+++ b/src/racer/cargo.rs
@@ -518,8 +518,32 @@ fn getstr(t: &toml::Table, k: &str) -> Option<String> {
 fn find_cargo_tomlfile<P>(file: P) -> Option<PathBuf>
     where P: Into<PathBuf>
 {
-    find_next_crate_root(file)
-        .map(|mut f| { f.push("Cargo.toml"); f })
+    find_next_crate_root(file).map(|mut f| {
+        f.push("Cargo.toml");
+        f
+    })
+}
+
+/// Find workspace root by traversing up the directory tree
+///
+/// Checks every Cargo.toml found while searching up the directory tree to see if it has a
+/// [workspace] section in its top level.
+fn find_workspace_root<P>(path: P) -> Option<PathBuf>
+    where P: Into<PathBuf>
+{
+    let mut path = path.into();
+    path.push("Cargo.toml");
+    trace!("find_workspace_root checking {:?}", path);
+    if path.exists() &&
+       parse_toml_file(&path).as_ref().and_then(|toml| toml.get("workspace")).is_some() {
+        trace!("find_workspace_root found workspace section in {:?}", path);
+        path.pop();
+        Some(path)
+    } else if path.pop() && path.pop() {
+        find_workspace_root(path)
+    } else {
+        None
+    }
 }
 
 /// Find crate root by traversing up the directory tree searching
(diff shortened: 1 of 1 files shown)


src/racer/cargo.rs
```diff
@@ -673,9 +697,18 @@ pub fn get_crate_file(kratename: &str, from_path: &Path) -> Option<PathBuf> {
...
+        // also search workspaces for a lockfile
         trace!("get_crate_file tomlfile is {:?}", tomlfile);
+
+        let absolute_path = env::current_dir().unwrap().join(from_path);
+        let workspace = find_workspace_root(absolute_path);
+
         let mut lockfile = tomlfile.clone();
         lockfile.pop();
+        if workspace.is_some() {
+            lockfile.push(workspace.unwrap());
```
jwilm: I'm confused what's happening here. It appears lockfile should have a value like `/home/foo/bar/baz/Cargo.lock`. Then, we push the absolute path to a workspace root on the end of it. Assuming `bar` is a workspace root, this would become something like `/home/foo/bar/baz/Cargo.clok/home/foo/bar`. Am I missing something?

src/racer/cargo.rs
```diff
@@ -673,9 +697,18 @@ pub fn get_crate_file(kratename: &str, from_path: &Path) -> Option<PathBuf> {
...
         // look in the lockfile first, if there is one
+        // also search workspaces for a lockfile
         trace!("get_crate_file tomlfile is {:?}", tomlfile);
+
+        let absolute_path = env::current_dir().unwrap().join(from_path);
+        let workspace = find_workspace_root(absolute_path);
+
         let mut lockfile = tomlfile.clone();
         lockfile.pop();
+        if workspace.is_some() {
```
jwilm: This could use the `if let Some(workspace) = workspace` construct to avoid the subsequent `unwrap`.

src/racer/cargo.rs
GoldenBadger: Unless I'm misunderstanding the docs (https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.push), since `workspace` is an absolute path, it should completely overwrite `lockfile`.

src/racer/cargo.rs
GoldenBadger: Sure thing, I'll tweak this when I'm home from work.

src/racer/cargo.rs
jwilm: Oh wow, that's kind of unexpected! TIL.

I wonder if we could use an assignment instead to make the intent more clear? I worry about others having the same confusion I did when reading the code.

GoldenBadger: Agreed, that would be much better for readability. I thought it was a strange case for `push` as well.

src/racer/cargo.rs
```diff
@@ -705,8 +705,8 @@ pub fn get_crate_file(kratename: &str, from_path: &Path) -> Option<PathBuf> {
 
         let mut lockfile = tomlfile.clone();
         lockfile.pop();
-        if workspace.is_some() {
-            lockfile.push(workspace.unwrap());
+        if let Some(workspace) = workspace {
+            lockfile = workspace;
```
TedDriggs: Given that this behavior was surprising to both you and @jwilm, could you please add a comment here explaining that (including the link you provided in the PR?)

src/racer/cargo.rs
```diff
@@ -518,8 +518,32 @@ fn getstr(t: &toml::Table, k: &str) -> Option<String> {
...
+    let mut path = path.into();
+    path.push("Cargo.toml");
+    trace!("find_workspace_root checking {:?}", path);
+    if path.exists() &&
+       parse_toml_file(&path).as_ref().and_then(|toml| toml.get("workspace")).is_some() {
+        trace!("find_workspace_root found workspace section in {:?}", path);
+        path.pop();
+        Some(path)
+    } else if path.pop() && path.pop() {
+        find_workspace_root(path)
```
TedDriggs: Can you add a comment here that mentions this is discarding `Cargo.toml`, moving up the directory tree, and trying again?

src/racer/cargo.rs
GoldenBadger: Is that necessary considering we won't be using `push` like that anymore?

GoldenBadger:
Hi, I would be happy to add an integration test, but the test project currently in `fixtures/` isn't suitable for testing workspaces.

PR #793, opened 2017-09-02
Adds support for inferring the type of a match expression

bkchr:
This patch only supports simple match expressions. For more complex match
expressions, the ExprTypeVisitor needs to be extended.

Fixes: https://github.com/racer-rust/racer/issues/755


TedDriggs:
Thanks for the PRs! I'm traveling this weekend but will aim to have feedback up by Wednesday.

PR #794, opened 2017-09-02
Rustfmt

bkchr:
It would be nice if the code would be formatted with rustfmt. :)


bkchr:
I used rustfmt 0.2.5-nightly. I also updated the commit message to include the version :)

PR #795, opened 2017-09-03
If let type inference

bkchr:
Requires: https://github.com/racer-rust/racer/pull/793

### DIFF
--- a/src/racer/scopes.rs
+++ b/src/racer/scopes.rs
@@ -366,17 +366,27 @@ fn expand_search_expr_handles_double_dot() {
     assert_eq!((2, 5), expand_search_expr("..foo", 4))
 }
 
+fn fill_gaps(buffer: &str, result: &mut String, start: usize, prev: usize) {
+    for _ in 0..((start-prev)/buffer.len()) { result.push_str(buffer); }
+    result.push_str(&buffer[..((start-prev)%buffer.len())]);
+}
+
 pub fn mask_comments(src: Src) -> String {
     let mut result = String::with_capacity(src.len());
     let buf_byte = &[b' '; 128];
     let buffer = from_utf8(buf_byte).unwrap();
     let mut prev: usize = 0;
     for (start, end) in src.chunk_indices() {
-        for _ in 0..((start-prev)/128) { result.push_str(buffer); }
-        result.push_str(&buffer[..((start-prev)%128)]);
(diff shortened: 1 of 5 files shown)


src/racer/scopes.rs
```diff
@@ -366,17 +366,27 @@ fn expand_search_expr_handles_double_dot() {
     assert_eq!((2, 5), expand_search_expr("..foo", 4))
 }
 
+fn fill_gaps(buffer: &str, result: &mut String, start: usize, prev: usize) {
```
TedDriggs: Is this change integral to the match/if-let work?

tests/system.rs
```diff
@@ -3982,3 +3982,127 @@ fn completes_for_global_path_in_trait_impl_decl() {
     assert_eq!(got.matchstr, "Bar");
     assert_eq!(got.mtype, MatchType::Trait);
 }
+
+// Issue #755
```
TedDriggs: Minor change; please put the full URL to the GitHub issue here.

src/racer/scopes.rs
bkchr: Yep, I changed the scope bounds for the IfLetVisitor and without these changes one test failed, because one point was out of bounds.

bkchr:
What do you mean by impacting the type in the match arm? It tries to get the return type of a match arm. ^^

TedDriggs:
```rust
let foo: Result<u32, String> = Ok(1);

match foo {
    Ok(val) => // ...,
    Err(err) => err.st~ // ... "arts_with"?
}
```

tests/system.rs
```diff
@@ -3982,3 +3982,127 @@ fn completes_for_global_path_in_trait_impl_decl() {
...
+        let f = File::open("hey");
+
+        let f = match f {
+            Err(error) =>  {
+                return "result".to_string();
+            },
+            Ok(file) => { file }
+        };
+
+        f.set_p~
```
TedDriggs: Please include the entire method call here, e.g. `f.set_p~ermissions(/* args */);` This comment applies to all tests.

bkchr:
> `let foo: Result<u32, String> = Ok(1); match foo { Ok(val) => // ..., Err(err) => err.st~ // ... "arts_with"? }`

For me that already works with the latest racer version^^

bkchr:
So, I addressed your comments :)

PR #797, opened 2017-09-19
More helpful error when rust-src is missing

mgsloan:
Note that this changes the behavior when "rustc --print sysroot" returns a
value. Before the change, if this command was successful, but the expected paths
not found in the yielded root, it would fallback on looking for the dir in
"/usr/...". I think this makes sense, if rustc is present and functional, it
should use its sysroot or fail. Otherwise, things are potentially inconsistent.

Also makes the following changes:

* Fixes environment variable name in description()

* Fixes double use of " around paths, display for PathBuf already includes
  quotes.

Note I am new to rust, so this change should probably be scrutinized carefully.  I did test it out, of course, it seems to work well.

### DIFF
--- a/src/racer/util.rs
+++ b/src/racer/util.rs
@@ -262,6 +262,7 @@ pub fn char_at(src: &str, i: usize) -> char {
 #[derive(Debug)]
 pub enum RustSrcPathError {
     Missing,
+    MissingRustSrc(path::PathBuf),
     DoesNotExist(path::PathBuf),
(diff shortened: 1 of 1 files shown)


TedDriggs:
This is showing all tests failing; is it dependent on another commit, or is it a WIP?

kngwyu:
Closing due to inactivity.
But, please feel free to reopen whenever you want.

PR #798, opened 2017-09-20
Make FileCache Send

tamird:
This is achieved by replacing a bunch of Rc with Arc, and by using lazy-init to make IndexedSource Sync, which makes Arc\<IndexedSource\> Send + Sync.

@nrc @jwilm @bkchr @TedDriggs

Closes #689.

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -105,7 +105,7 @@ impl Match {
     /// but in the interest of minimizing the crate's public API surface it's exposed
     /// as a private method for now.
     fn is_same_as(&self, other: &Match) -> bool {
-        self.point == other.point 
+        self.point == other.point
         && self.matchstr == other.matchstr
         && self.filepath == other.filepath
     }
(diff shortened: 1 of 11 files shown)


tamird:
Actually, hold off on this; there are bugs here.

tamird:
Alright, added some commits - will squash the last two before merging.

src/racer/core.rs
```diff
@@ -371,7 +372,7 @@ impl fmt::Debug for PathSearch {
 pub struct IndexedSource {
     pub code: String,
     pub idx: Vec<SourceByteRange>,
-    pub lines: RefCell<Vec<SourceByteRange>>
+    pub lines: Arc<RwLock<Vec<SourceByteRange>>>,
```
jwilm: Why is the `Arc` needed here?

src/racer/core.rs
```diff
@@ -412,21 +413,25 @@ impl IndexedSource {
...
 
     fn cache_lineoffsets(&self) {
-        if self.lines.borrow().len() == 0 {
-            let mut result = Vec::new();
-            let mut point = 0;
-            for line in self.code.split('\n') {
-                result.push((point, line.len() + 1));
-                point += line.len() + 1;
+        if self.lines.read().unwrap().len() == 0 {
+            let mut lines = self.lines.write().unwrap();
```
jwilm: Is the assumption with all of the `unwrap` calls that poisoning will not occur? It seems like poisoning is a real risk in case of a panic while holding the lock, but there's not much risk of corruption even in such a case.

Should we instead handle poisoning with `.unwrap_or_else(|e| e.into_inner())`?

src/racer/core.rs
tamird: Good point, and something I realized as well.

We really just want lazy initialization here - I'm looking for a suitable crate.

jwilm: What about just having a RwLock around an empty Vec?

tamird:
Hm, I'm not exactly sure how this enables https://github.com/rust-lang-nursery/rls/issues/33 - it seems that `vfs::Vfs` now needs to be Sync as well, to allow `Arc<Vfs>` to implement `FileLoader`. @nrc - any guidance?

PR #799, opened 2017-09-23
Racer now uses rustup to find the stdlib source code even when used as a library.

golddranks:
Refactored the `stdlib` source code search a bit: It now stores the path in a `lazy_static`, and doesn't load it directly from an env var at usage sites. When initializing the `lazy_static`, it checks both the `RUST_SRC_PATH` env var and `rustup`'s `rust-src` component. It used use `rustup` only when using Racer as a binary. Now it does it even when used as a library.

This allows using RLS without setting `RUST_SRC_PATH`; it suffices that `rustup` is installed. Before, it just used to crash if `RUST_SRC_PATH` wasn't set.

Also got rid of the warnings: converted a few `\\\` doc comments that weren't actual item doc comments to `\\` normal comments, and got rid of the non-used `extern crates` that were likely left from when the binary and library were not separated.


golddranks:
This helps with, although doesn't eliminate this: https://github.com/racer-rust/racer/issues/751

golddranks:
Oops, tests fail because the `env_logger` was still needed in tests. Fixing!

src/racer/config.rs
```diff
@@ -0,0 +1,23 @@
+use std::path::PathBuf;
```
TedDriggs: Having a generic `config.rs` makes me suspicious; it seems likely that it'll become a parking lot for a bunch of unrelated items. Can you make the name for this more specific to file paths?

src/racer/core.rs
```diff
@@ -997,9 +997,9 @@ fn complete_from_file_(
             let line = src_text[linestart..pos].trim().rsplit(';').nth(0).unwrap();
             debug!("Complete path with line: {:?}", line);
 
-            /// Test if the **path expression** starts with `::`, in which case the path
-            /// should be checked against the global namespace rather than the items currently
-            /// in scope.
+            // Test if the **path expression** starts with `::`, in which case the path
```
TedDriggs: Thanks for cleaning these up; it makes me sad to lose the ability to add doc comments to locals and have them show up in completions, but the rustdoc team has spoken.

src/racer/nameres.rs
```diff
@@ -11,13 +11,9 @@ use matchers::find_doc;
 use cargo;
 use std::path::{Path, PathBuf};
 use std::{self, vec};
+use std::iter;
```
TedDriggs: Can you combine this with the `use` on the previous line?

src/racer/nameres.rs
```diff
@@ -1091,15 +1084,12 @@ pub fn search_prelude_file(pathseg: &core::PathSegment, search_type: SearchType,
     let mut out : Vec<Match> = Vec::new();
 
     // find the prelude file from the search path and scan it
-    let srcpaths = match std::env::var("RUST_SRC_PATH") {
-        Ok(paths) => paths,
-        Err(_) => return out.into_iter()
+    if config::RUST_SRC_PATHS.is_empty() {
```
TedDriggs: This change makes the comment above inaccurate; can you please update it?

src/racer/config.rs
golddranks: Yes, I can; however, I originally thought that it would be actually good to have a centralised module for reading any needed configuration from the environment and keeping the state. Do you think it would make sense to move this to `nameres.rs` where `RUST_SRC_PATHS` is used?

src/racer/core.rs
golddranks: Did they actually consider that use case? It might be that they just missed it, and deprecated this as an useless thing.

src/racer/nameres.rs
golddranks: Aye.

src/racer/nameres.rs
golddranks: Will do.

golddranks:
What is the purpose of the fixtures directory?

golddranks:
Ping @TedDriggs , in case you didn't notice my updates and question.

src/racer/core.rs
TedDriggs: They considered it; I asked them about it, which was what prompted the discussion which ended in deprecation.

src/racer/config.rs
TedDriggs: I'm personally a fan of smaller, purpose-oriented modules, so I'd suggest one that only contained this and have that be used by `nameres`. However, I'm fine with it living in `nameres.rs` directly for the time-being.

TedDriggs:
> What is the purpose of the fixtures directory?

I actually don't know, but I got the impression that it was supposed to do something around environment tests like this.

jwilm:
The fixtures directory has projects for tests. It allows us to test very specific setups; although, it hasn't gotten much use outside of 1-2 tests.

src/racer/config.rs
golddranks: All right, I'll refactor that.

golddranks:
Since there was a TODO comment that indicated that a more thorough refactoring might be good, I went ahead and did that.

It now resolves to a single path using multiple sources (the env var, rustup and default path list) that is verified to contain the stdlib source.

Also added tests.

golddranks:
I'd very much appreciate if, when this gets pulled in, a new version of Racer would be released; updating RLS to use the new version, it would make RLS setup one step easier.

golddranks:
The Windows build fails the test that uses `rustup` to get the path. Is rustup present on Windows? And does it work similarly than on macOS, which is the system I'm using?

TedDriggs:
The Windows build seems to work on all but one architecture, so I don't think it's a problem with the AppVeyor config.

golddranks:
All right, all requested changes are done (Although Github claims otherwise) and tests are passing.

golddranks:
Friendly ping to @TedDriggs . This PR is waiting for a (hopefylly) final review for merge.

src/racer/util.rs
```diff
@@ -351,54 +362,142 @@ fn check_rust_sysroot() -> Option<path::PathBuf> {
...
+
+    let _guard = TEST_SEMAPHORE.lock().unwrap();
+
+    let original = env::var_os("RUST_SRC_PATH");
+    if env::var_os("RUST_SRC_PATH").is_none() {
+        env::set_var("RUST_SRC_PATH", check_rust_sysroot().unwrap());
+    }
+    assert!(get_rust_src_path().is_ok());
+
+    match original {
```
bkchr: Nitpick: You could create a struct, that stores the original RUST_SRC_PATH and implement Drop. If drop is called, you would reset to the original env var. This could reduce some code duplication in the tests :)

golddranks:
Ping @jwilm , is there anything I can do to make this go forward?

golddranks:
@TedDriggs is there any other maintainers that could approve of this?

PR #807, opened 2017-10-31
Fixing a bug that makes 2.0.11 unusable

golddranks:
There was a bug I didn't notice in my refactoring of the path search code. This PR fixes it, and adds a test case that should catch in the future if racer stops finding matches in some basic cases.

Sorry for the blunder. 2.0.11 should be yanked and 2.0.12 released with a fast schedule.

Also made cleaned up earlier tests in `nameres.rs` so that they won't block other tests from running if a single one fails.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,18 +1,3 @@
-[root]
-name = "racer"
-version = "2.0.10"
-dependencies = [
- "clap 2.26.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


golddranks:
Fixes https://github.com/racer-rust/racer/issues/806

golddranks:
Ping @TedDriggs @jwilm

jwilm:
Thanks @golddranks; will do.

jwilm:
Yanked 2.0.11 and published 2.0.12.

PR #808, opened 2017-11-09
Made RUST_SRC_PATH optional to support finding local files even when std source is missing.

golddranks:
This fixes another regression from 2.0.10: Version 2.0.12 panics right away if it doesn't find the std source. This prevents searching for local files even in absence of the std source, which used to be possible.

It now stores the std source path as an optional value instead of panicking.

The regression made the RLS CI to break, because they didn't have the std source path installed there. They now have, but not panicking when std isn't found might still be a valuable thing, plus regressions aren't nice.

Also added a test case that searched for a local file. (Added a submodule to `fixtures/arst` for that.)

### DIFF
--- a/fixtures/arst/src/submodule/mod.rs
+++ b/fixtures/arst/src/submodule/mod.rs
@@ -1,0 +1,3 @@
+pub fn hello_submodule() {
+	println!("Hello from submodule.");
+}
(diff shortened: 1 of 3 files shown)


golddranks:
Ping @TedDriggs @jwilm , this, too addresses a regression found when using racer with RLS.

kngwyu:
@golddranks 
Could you rebase and add some `warn!` for the case if we can't get `RUST_SRC_PATH`?

golddranks:
Done! Waiting for the CI.

kngwyu:
Thanks!
Appveyor's failing has nothing to do with this PR, so please don't mind.

PR #810, opened 2017-11-20
Include the type name with snippets

sebastiencs:
In addition to the variable name, snippets now include their type.
IMHO, the type is more important to know that only the variable name.


kngwyu:
Thanks for PR, and sorry for long delay of review.
@sebastiencs 
Is is breaking change for clients?
If so, I think it's better to implement another command.

sebastiencs:
@kngwyu  no problem, thanks for your work on racer.
It doesn't break anything. The clients will just show more informations about the parameters

kngwyu:
OK, then it's good to add type name.
Thanks!

Issue #816, opened 2018-01-16
Racer fails to complete with a multibyte comment in a match statement 


kngwyu:
When a comment with multibyte chars is inserted where match expressions should be placed, racer fails to complete in the area between that comment and the end of the match statement.
See below.
```rust
fn main() {
    let a = 3;
    let long_name_variable = 0;
    let s = match a {
        1 => "cat",
        2 => "ls",
        // comment with multibyte chars, like ★
        _ => {
            let b = long_name_v..; // this fails
            "wc"
        }
    };
    let c = long_name_v..; // this succeeds
}
```
In this example,  racer fails to complete ```let b = long_name_variable```, but succeeds ```let c = long_name_variable```.
I'm using racer 2.0.12.

TedDriggs:
Fixed in #819

PR #817, opened 2018-01-16
Solved Issue 816

kngwyu:
Solved #816.
Added check if the byte is newline byte, before calling ```util::is_pattern_char(char_at(src, i))```.
It's because multibyte char can occur over '\n' (in case of 1-liner comment).
And I found another problem around the  ```get_start_of_pattern```  so I will add new issue(#818).


kngwyu:
I found this code fails in cases like below.
```rust
fn main() {
    let option = Some(5);
    let _ = match option {
        Some(variable)
            => {
            let b = variable;
            3
        }
        None => 4,
    };
}
```

Issue #818, opened 2018-01-16
Racer can't recognize the variable binded by pattern matching if C-style comment is inserted


kngwyu:
See below.
```rust
fn main() {
    let option = Some(5);
    let _ = match option {
        Some(variable) /* comment */=> {
            let b = vari~;
            3
        }
        None => 4,
    };
}
```
Racer can't complete ```let b = variable```.

TedDriggs:
Fixed in #819

PR #819, opened 2018-01-16
ignore comments when finding the start of pattern

kngwyu:
For #816 and #818.

### DIFF
--- a/src/racer/util.rs
+++ b/src/racer/util.rs
@@ -27,6 +27,114 @@ pub fn is_ident_char(c: char) -> bool {
     c.is_alphanumeric() || (c == '_') || (c == '!')
 }
 
+// to ignore C-Style Comment
+#[derive(Debug)]
+enum CommentAutomatonState {
+    None,
+    Start, // just / occurs
+    InComment, // /* ...
+    BeforeEnd, // /* ... *
+    InSlash, // like ///// (maybe not useless)
+}
+
+impl CommentAutomatonState {
+    // in case we can know wer'e in comment without looking ahead
+    fn is_in_comment(&self) -> bool {
+        match *self {
+            CommentAutomatonState::InComment
+            | CommentAutomatonState::BeforeEnd
+            | CommentAutomatonState::InSlash => true,
+            _ => false,
+        }
+    }
+}
+
+impl Default for CommentAutomatonState {
+    fn default() -> CommentAutomatonState {
+        CommentAutomatonState::None
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct CommentAutomaton {
+    state: CommentAutomatonState,
+}
+
+impl CommentAutomaton {
+    pub fn cur_state(&self) -> bool {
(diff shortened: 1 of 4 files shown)


src/racer/util.rs
```diff
@@ -23,6 +27,114 @@ pub fn is_ident_char(c: char) -> bool {
...
+    }
+}
+
+impl Default for CommentAutomatonState {
+    fn default() -> CommentAutomatonState {
+        CommentAutomatonState::None
+    }
+}
+
+#[derive(Debug, Default)]
```
TedDriggs: Please add a rustdoc comment explaining what this is for.

src/racer/util.rs
```diff
@@ -23,6 +27,114 @@ pub fn is_ident_char(c: char) -> bool {
...
+pub struct CommentAutomaton {
+    state: CommentAutomatonState,
+}
+
+impl CommentAutomaton {
+    pub fn cur_state(&self) -> bool {
```
TedDriggs: Please rename to `is_in_comment`

src/racer/util.rs
```diff
@@ -23,6 +27,114 @@ pub fn is_ident_char(c: char) -> bool {
...
+enum CommentAutomatonState {
+    None,
+    Start, // just / occurs
+    InComment, // /* ...
+    BeforeEnd, // /* ... *
+    InSlash, // like ///// (maybe not useless)
+}
+
+impl CommentAutomatonState {
+    // in case we can know wer'e in comment without looking ahead
```
TedDriggs: I don't understand this comment; or maybe it's that the method name doesn't match the comment?

Typo: Should be "we're"

src/racer/util.rs
```diff
@@ -23,6 +27,114 @@ pub fn is_ident_char(c: char) -> bool {
...
 }
 
+// to ignore C-Style Comment
+#[derive(Debug)]
+enum CommentAutomatonState {
+    None,
+    Start, // just / occurs
+    InComment, // /* ...
+    BeforeEnd, // /* ... *
+    InSlash, // like ///// (maybe not useless)
```
TedDriggs: Please use rustdoc comments for each enum variant, rather than postfixed comments.

kngwyu:
Thanks for the review, and I came up with a reimplemention of  ```CommentAutomaton```  to iterator, such as ```CommentSkipIterRev``` for easiness to read.
Do you think it's better?

kngwyu:
Updated(packed Comment-Skip logic into ```CommentSkipIterRev``` in codecleaner.rs and add some tests).
And the test which fails isn't related to my commits so I created another PR(#820).

TedDriggs:
> In addition, I used rustfmt to system.rs by mistake... (Should I fix this?)

Yes please. `rustfmt` is good, but it'll make this much harder to review. FYI, I'm going to try and review this early next week.

kngwyu:
Thanks for advice.
I fixed and rebased.

Issue #785, opened 2017-08-08
Racer shouldn't need type annotations to understand enum variants


TedDriggs:
Repro code:

```rust
fn foo() {
    let bar = Some("Hello");
    bar.unwrap_or_d~efault();
}
```

**Expected:** Completion at the cursor should find `unwrap_or_default`
**Observed:** Completion finds nothing, because it "can't find the type of an EnumVariant"

The least invasive fix seems to be identifying the call to the enum variant and resolving its parent enum type.

kngwyu:
Closed via #825

PR #825, opened 2018-02-08
Issue 785

kngwyu:
Solves #785.
When ```resolve_ast_path```returns a Match with ```EnumVariant```, racer already detects the name of enum.
So I changed ```resolve_path``` and ```MatchType```, so that if it detect the ```EnumVariant``` it returns  Match with ```EnumVariant```  and Match with ```Enum```.


nrc:
@kngwyu could you rebase this PR please? Then I think it looks good to merge.

kngwyu:
Rebased.
The reason why test fails is [here](https://github.com/kngwyu/racer-nightly/commit/6df7414345a4ca1bd6ab1619957fd8bfbd273339)

nrc:
Thanks!

Issue #826, opened 2018-02-16
Support //! style comment as docstring for crate


lbolla:
See discussion in #566.

For crates whose docstrings are in `lib.rs`, Racer's `complete-with-snippet` returns an empty list when completing the crate name itself. Compare with crates which instead use `mod.rs`, which behave correctly.

E.g. `env_logger`

kngwyu:
Closed via #827

PR #827, opened 2018-02-16
add //! style comment support as docstring for crates

kngwyu:
Solves #826.

### DIFF
--- a/src/test_project/test_fixtures/src/lib.rs
+++ b/src/test_project/test_fixtures/src/lib.rs
@@ -1,3 +1,4 @@
+//! This is a test project for racer.
 
 pub mod foo;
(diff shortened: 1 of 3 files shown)


src/test_project/test_fixtures/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+//! This is a test project for racer.
```
TedDriggs: We've had issues before with multiline comments that include paragraph-separator lines. Given that crate docs are disproportionately likely to be long and have code samples, can you flesh out this comment to exercise:

* Paragraph break
* Code block, including a doc comment like this one inside that example block?

kngwyu:
Thanks for review.
I added more doc comments.

Issue #706, opened 2017-04-05
Unable to autocomplete based on trait bounds


JonasZ95:
```rust
trait X {
    fn do_x(&self);
}

struct A ();
struct B<T> {
   x: T
}

impl X for A {
    fn do_x(&self) {
        println!("A");
    }
}

impl<T> B<T> where T: X {
     fn do_b(&self) {
         self.x.//No autocomplete
    } 
}
```

That's what I get(:
```
racer complete 18 17 main.rs
PREFIX 204,204,
END
```

I use racer 2.0.6. May be a duplicate but I was unable to find something similar.

kngwyu:
Closed in favour of #939

PR #828, opened 2018-02-18
Completion by trait bounds for fnarg

kngwyu:
By this PR, racer can complete method for generic type with trait bounds, like
```rust
fn f<T: Clone>(a: T) {
    let b = a.clo~
}
```
In addition, trait inheritance(like below) is also supported.
```rust
trait Tr: Clone {}
fn f<T: Tr>(a: T) {
    let b = a.clo~
}
```

This PR is related to #706, but what I implemented in this PR works only when ```FnArg``` has generic type `T`.
So, these cases are not supported now.
```rust
struct S<T> {
    mem: T
}
fn f<T: Clone>(a: S<T>) {
    let b = a.mem.clo~
}
impl<T: Clone> S<T> {
    fn f(self) {
        let b = a.mem.clo~
    } 
}
```
I'm going to implement these features after this PR and #825  are merged.


kngwyu:
This change isn't compatible with current `libsyntax` so I close this.

PR #838, opened 2018-04-04
Cache parsed TOML and cargo crate roots - 10x speed for standalone use

birkenfeld:
This reduces the number of toml parsings to 226 for completing "std::".
Before it was over 9000! (insert meme here).

I also made a change to logging so that it is more suitable for checking the timings of individual steps.

PR also includes a few minor things I found on the way - best to review each commit independently.

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -813,7 +813,7 @@ impl<'c> Session<'c> {
     /// 1. overrides
     /// 2. lock file
     /// 3. toml file
-    pub fn get_crate_file(&self, kratename: &str, from_path: &path::Path) -> Option<path::PathBuf> {
+    pub(crate) fn get_crate_file(&self, kratename: &str, from_path: &path::Path) -> Option<path::PathBuf> {
         self.cargo_roots.borrow_mut().entry((kratename.into(), from_path.into()))
             .or_insert_with(|| cargo::get_crate_file(kratename, from_path))
             .clone()
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -18,6 +18,7 @@ use nameres;
 use ast;
 use codecleaner;
 use util;
+use cargo;
 
 /// Within a [`Match`], specifies what was matched
 ///
@@ -743,6 +744,8 @@ pub struct Session<'c> {
     pub generic_impls: RefCell<HashMap<(path::PathBuf, usize),
                                        Rc<Vec<(usize, String,
                                                ast::GenericsVisitor, ast::ImplVisitor)>>>>,
+    /// Cache for crate roots
+    cargo_roots: RefCell<HashMap<(String, path::PathBuf), Option<path::PathBuf>>>,
 }
 
 impl<'c> fmt::Debug for Session<'c> {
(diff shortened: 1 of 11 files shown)


birkenfeld:
Right, some invalidation is needed for that case. I'll implement something.

src/racer/core.rs
```diff
@@ -796,6 +800,18 @@ impl<'c> Session<'c> {
...
     }
 
+    /// Find the library root for kratename
+    ///
+    /// The library is searched for by checking
+    ///
+    /// 1. overrides
+    /// 2. lock file
+    /// 3. toml file
+    pub fn get_crate_file(&self, kratename: &str, from_path: &path::Path) -> Option<path::PathBuf> {
```
jwilm: What do you think about making this `pub(crate)` instead? With the 2.0 release, we dramatically reduced the public API surface. At this point, I'd like to take a conservative approach with expanding that API at all.

src/racer/core.rs
```diff
@@ -738,6 +739,8 @@ pub struct Session<'c> {
     /// The file cache is used within a session to prevent multiple reads. It is
     /// borrowed here in order to support reuse across Racer operations.
     cache: &'c FileCache,
+    /// Cache for crate roots
+    cargo_roots: RefCell<HashMap<(String, path::PathBuf), Option<path::PathBuf>>>,
```
jwilm: There is interest in having the `Session` type be sync. Are you open to changing this to a `RwLock` instead?

src/racer/core.rs
birkenfeld: Sure, still have to get used to `pub(crate)`.

src/racer/core.rs
birkenfeld: That's your call of course. (Maybe adding a test would be good to keep it `Sync`?)

Hmm, if Racer stuff is run in different threads, it might also be better to make the TOML cache an `Arc<Mutex>` lazy-static instead of a thread local?

nrc:
@birkenfeld would you be able to rebase and address the last few review comments please? It looks like this is ready to merge then.

birkenfeld:
I will once I'm back from vacation :)

src/racer/core.rs
birkenfeld: This is now done.

src/racer/core.rs
birkenfeld: `Session` is currently not `Sync` (and hasn't been for a long time, because `FileCache` isn't). If you want to make that change, I'd propose to do it separately.

nrc:
Thanks!

birkenfeld:
Can we make a new release with these changes?

nrc:
Published 2.0.14

kngwyu:
Here's AppVeyour's log 
```
failures:
    completes_pub_fn_from_local_package
    completes_pub_fn_from_local_submodule_package
    find_crate_doc
    finds_extern_crate
    follows_use_local_package
test result: FAILED. 177 passed; 5 failed; 5 ignored; 0 measured; 0 filtered out
```

these 5 tests are for outer crates so maybe related in this PR.
@birkenfeld 
Do you have any idea?

But honestly I don't think we have to take it seriously. I'm planning migrate https://github.com/kngwyu/racer-nightly/pull/3 to upstream, then we do tests for outer crates in a bit different way/

PR #840, opened 2018-04-07
Revamp of the cargo module

birkenfeld:
(based on the cache_tomls branch - so beware, seemingly unrelated commits)

This is a general overhaul of the code that removes a few unwraps and enforces consistent naming, but shouldn't change the logic. It will be expanded to fix a few issues with unconventional crate setups later.


kngwyu:
Looks overlapping with #838, so I close this.
But feel free to reopen PR.

PR #848, opened 2018-05-11
Remove Chain<Chain<Chain<Chain ...

jwilm:
Thanks, impl Trait!


kngwyu:
Looks fine for me.
But now by https://github.com/rust-lang/rust/pull/49896, some string methods like `starts_with` are declared in macro so we have some broken tests.
So let's merge this PR after fix them.

kngwyu:
Thanks！

Issue #624, opened 2016-09-28
Cross-crate reexports are not completed


bluss:
For example, itertools reexports: `pub use either::Either;` but `Either` is not completable under `itertools::` when completing it from cargo sources (when from a `path` dependency, it is ok).

Another example, generic-array reexports `pub extern crate typenum;` which is not available under `generic_array::`.

bluss:
What's more absurd is that it offers to complete `generic_array::nodrop` and `generic_array::std` which are precisely its private `extern crate` lines?? But not the public extern crate line.

PR #850, opened 2018-05-12
Skip pub keyword when parsing extern crate

cellscape:
This change allows racer to follow and parse reexported crates in
external crates.

For example, now it's possible to go to definition of
`amethyst::renderer` which is reexported in `amethyst` as `pub extern
crate amethyst_renderer as renderer`.

#624

### DIFF
--- a/src/racer/matchers.rs
+++ b/src/racer/matchers.rs
@@ -43,6 +43,39 @@ pub fn match_values(src: Src, blobstart: Point, blobend: Point,
     it.chain(match_macro(&src, blobstart, blobend, searchstr, filepath, search_type, local).into_iter())
 }
 
+fn strip_keyword_prefix(src: &str, keyword: &str) -> Option<Point> {
+    let mut start = 0usize;
+    if src.starts_with(keyword) {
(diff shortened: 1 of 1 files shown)


kngwyu:
Thanks for PR!

But regarding an `extern crate~` sentence as a public module is racer's bug (see also https://github.com/kngwyu/racer-nightly/pull/4).
So, to handle crate visibility correctly, we have to change the default behavior for  `extern crate ~` statement.
However, this kind of fix is not so easy, so to skip `pub` is not so bad as an immediate fix.

@nrc 
What do you think of this kind of PR?

cellscape:
> But regarding an extern crate~ sentence as a public module is racer's bug (see also kngwyu#4).
So, to handle crate visibility correctly, we have to change the default behavior for extern crate ~ statement.

Yeah I agree that visibility should be handled correctly. However I think even if proposed change doesn't fix visibility issues it still improves things for now, doesn't seem to break anything and helps already with at least some use cases.

kngwyu:
Looks good.
Thanks!

PR #854, opened 2018-05-18
 Replace syntex with rustc-ap-syntax 

kngwyu:
Now `syntex_syntax` is no longer maintained, so we have to move towards `libsyntax`.

Here I list up change from our current `syntex_syntax 52.0` and `rustc-ap-syntax 133.0` (though not all).
- `visit::Visitor` needs life time(now its declaration is `pub trait Visitor<'ast>: Sized`)
- data structure for `use` is completely different
- `ExprKind::TupField` is removed by https://github.com/rust-lang/rust/pull/49718
- we have to call [scoped-tls](https://github.com/alexcrichton/scoped-tls) explicitly to get names of idents
- and some changes..

By this change, we become unable to compile racer with `beta` or `stable` compiler.
So we have to decide some things.

- Should we change the package name like `rustfmt-nightly`?
- Should we remain current master as `syntex` branch?

I constructed this PR from https://github.com/kngwyu/racer-nightly/pull/1 https://github.com/kngwyu/racer-nightly/pull/2 https://github.com/kngwyu/racer-nightly/pull/5/commits/40466f66f50fd28fa8e845107d62a30448d113e3 and some commits.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,19 +2,19 @@
 environment:
   matrix:
     # Stable channel
-    - TARGET: i686-pc-windows-gnu
-      CHANNEL: stable
-    - TARGET: i686-pc-windows-msvc
-      CHANNEL: stable
-    - TARGET: x86_64-pc-windows-gnu
-      CHANNEL: stable
-    - TARGET: x86_64-pc-windows-msvc
-      CHANNEL: stable
+    # - TARGET: i686-pc-windows-gnu
+    #   CHANNEL: stable
+    # - TARGET: i686-pc-windows-msvc
+    #   CHANNEL: stable
+    # - TARGET: x86_64-pc-windows-gnu
+    #   CHANNEL: stable
+    # - TARGET: x86_64-pc-windows-msvc
+    #   CHANNEL: stable
     # Beta channel
-    - TARGET: i686-pc-windows-gnu
-      CHANNEL: beta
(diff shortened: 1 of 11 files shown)


birkenfeld:
Why is nightly required?

kngwyu:
@birkenfeld 
Because rustc's libsyntax uses [many feature gates](https://github.com/rust-lang/rust/blob/master/src/libsyntax/lib.rs#L21-#L28)

davemilter:
@kngwyu 

Why rustc-ap-syntax and not https://github.com/dtolnay/syn ?

kngwyu:
@davemilter 
For some reasons.
- We currently use `syntex-syntax` i.e. older version of libsyntax. So to migrate over to `rusc-ap-syntax` i.e. latest version of libsyntax is not so hard.
- `syn` is designed for `proc_macro`
But, if you think there's any merit to use `syn`, please let me know.

nrc:
`syn` does have some advantages - it works on stable and is itself more stable (changes less, so less bustage for clients). However, I think moving to syn will be a really big change and using rustc_ap_syntax is a great step for right now (in particular, rustfmt uses it, so at the least we have a couple of important tools relying on the same dependency).

nrc:
Thanks, this is great!

kngwyu:
I created `syntex` branch for just in case.
And, we have to update README before next release.

PR #859, opened 2018-05-27
matchers: remove a pretty useless debug info!()

birkenfeld:


kngwyu:
I think it's completely useless, too.
Thanks!

PR #860, opened 2018-05-27
Run clippy and fix most warnings.

birkenfeld:


kngwyu:
Looks nice to me, but I have one thing to worry about:
Isn't our clippy too old?
I think it's better for us to update our clippy in this PR.

birkenfeld:
Hmm, I didn't even notice that. Do we need clippy in the dependencies anymore? I just ran `cargo clippy`, which works independently.

I pushed another change with the dependency removed. To get the custom options for lint "deny", you'll have to run `cargo clippy --features clippy`.

kngwyu:
> I pushed another change with the dependency removed. 

I agree with this change: to have clippy as dependency forces us to update it frequently  so `cargo-clippy` is more convenient.

PR #863, opened 2018-06-01
Readme update: RUST_SRC_PATH not mandatory anymore

apiraino:
Hello,

as briefly discussed [here](https://github.com/racer-rust/emacs-racer/issues/93#issuecomment-393747999) the env var `RUST_SRC_PATH` is to not mandatory anymore.

If this env var is actually not really needed, could it be useful to relax the configuration requirements for `racer`? Possibly this may lead to a couple of `racer` issue disappearing (not really sure about this, haven't investigated really closely).

And how about removing mentioning this env var altogether to simplify the config process?

Feel free to suggest on how to improve this little doc update or OTOH is not really needed and can be yanked.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,10 +36,14 @@ As mentioned in the command output, don't forget to add the installation directo
 
     2. manually from git, or download from https://www.rust-lang.org/install.html (the 'rustc' source download behind the 'source' link is the right one).
 
-2. Set the ```RUST_SRC_PATH``` environment variable to point to the 'src' dir in the Rust source installation
+    Ensure to download sources for any rust version you have installed (e.g. a common configuration is to have the `stable` and many `nightly` installed).
+
+2. (OPTIONAL) Set the ```RUST_SRC_PATH``` environment variable to point to the 'src' dir in the Rust source installation
 
    (e.g. ```% export RUST_SRC_PATH=/usr/local/src/rust/src``` or ```% export RUST_SRC_PATH="$(rustc --print sysroot)/lib/rustlib/src/rust/src"``` )
-   
+
+   This is not mandatory anymore to use `racer`.
+
 3. Test on the command line:
 
    ```racer complete std::io::B ```  (should show some completions)
--- a/README.md
+++ b/README.md
@@ -88,10 +88,10 @@ Vim integration has been moved to a separate project: [vim-racer](https://github
 
 Racer recommends the [`vscode-rust` extension](https://github.com/editor-rs/vscode-rust). This is an actively-maintained fork of the now-deprecated [`RustyCode` extension](https://github.com/saviorisdead/RustyCode).
 
-### Atom integration 
+### Atom integration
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,10 +36,14 @@ As mentioned in the command output, don't forget to add the installation directo
 
     2. manually from git, or download from https://www.rust-lang.org/install.html (the 'rustc' source download behind the 'source' link is the right one).
```
kngwyu: Now we can't install source from `https://www.rust-lang.org/install.html`, so could you please modify this sentence to just 'manually from [git](https://github.com/rust-lang/rust.git)' ?

README.md
```diff
@@ -36,10 +36,14 @@ As mentioned in the command output, don't forget to add the installation directo
...
 
-2. Set the ```RUST_SRC_PATH``` environment variable to point to the 'src' dir in the Rust source installation
+    Ensure to download sources for any rust version you have installed (e.g. a common configuration is to have the `stable` and many `nightly` installed).
```
kngwyu: > to have the `stable` and many `nightly` installed

I don't understand what you mean here.
But, I'm not a native English speaker so I'm not sure.

apiraino: neither am I :-) what I tried to explain here is perphaps redundant but Ii happened to me yesterday; imagine the following workflow:

- user downloads stable rust using `rustup`
- user downloads sources and installs `racer`
- at some point in the future, user downloads a nightly build because (f.e.) they want to use Rocket 
- `racer` partially stops working because user forgot to also download the nightly build sources

([just happened to me yesterday](https://github.com/racer-rust/emacs-racer/issues/93#issuecomment-393495176))

Do you think it's worth such a note of it ends up being more confusing? Any suggestion on how to rephrase more clearly? e.g. simply with `Remember to download sources after installing a nightly build too`

README.md
apiraino: ack

README.md
```diff
@@ -36,10 +36,14 @@ As mentioned in the command output, don't forget to add the installation directo
...
+
+2. (OPTIONAL) Set the ```RUST_SRC_PATH``` environment variable to point to the 'src' dir in the Rust source installation
 
    (e.g. ```% export RUST_SRC_PATH=/usr/local/src/rust/src``` or ```% export RUST_SRC_PATH="$(rustc --print sysroot)/lib/rustlib/src/rust/src"``` )
-   
+
+   This is not mandatory anymore to use `racer`.
```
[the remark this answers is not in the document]
apiraino: ack

README.md
kngwyu: Ah I understand, thanks.
Then my suggestion is:
**Note** 
If you want to use racer with multiple release channels(we can use 3 release channels, stable, beta and nightly), you have to download Rust sourcecode for each relase channels.
e.g.(rustup case)
You have to run `rustup component add rust-src` after `rustup toolchain add nightly`.

apiraino: ack, thank you!

kngwyu:
Thanks!

PR #866, opened 2018-06-05
Add better handling of relative path dependencies

aloucks:
Autocomplete now works with path dependencies such as "./my-dep" and "../my-dep"
on Windows.


aloucks:
Ah, just noticed that https://github.com/racer-rust/racer/pull/855 was merged.

Issue #844, opened 2018-04-26
Slow behaviour with glob imports


mvr:
(This could be a non-issue, there's a good chance I'm using `use` wrong.) 

Racer takes 7 seconds on my machine to provide completions in the following situation, even when none of the modules `a` through `i` exist. If you uncomment the last `use`, it takes around a minute.

```rust
use a::*;
use b::*;
use c::*;
use d::*;
use e::*;
use f::*;
use g::*;
use h::*;
// use i::*;

pub fn foo() -> () {
    Whatever::
}
```

kngwyu:
It's interesting and very annoying.
Thanks for reporting.

kngwyu:
Maybe it's because [search_prelude_file](https://github.com/racer-rust/racer/blob/master/src/racer/nameres.rs) runs many times and it seems difficult to fix.
We can accelerate processing by running search_prelude_file just once using regex, but for it, we have to change many codes.

mvr:
When I run with RUST_LOG=debug, it looks like there is something strange and exponential going on in `match_use`: There are many lines like

```
DEBUG 2018-04-27T02:37:13Z: racer::matchers: found a glob: now searching for P[a::c]
DEBUG 2018-04-27T02:37:13Z: racer::matchers: found a glob: now searching for P[d::a]
DEBUG 2018-04-27T02:37:13Z: racer::matchers: found a glob: now searching for P[d::c]
```

in various combinations of `a`-`h`.

kngwyu:
@mvr 
Thanks.
That's interesting.

kngwyu:
Ah I understand.
When racer tries to find the definition of `a`, it searches `b::a` for the situation like this.
```rust
mod b {
    pub mod a {}
}
use a::*;
use b::*;
```

So I think we can reduce this kind of complexity by caching the result of `resolve_path`.

kngwyu:
Closed via #875 (though it's still a bit slow)

mvr:
Hopefully this was the problem with #803 and #757 too!

PR #875, opened 2018-06-25
issue-844

kngwyu:
For #844 
Now what I did is only skip glob after we construct a path from glob and 2 tests fail.
I'll fix tests by this weekend.

### DIFF
--- a/src/racer/matchers.rs
+++ b/src/racer/matchers.rs
@@ -12,13 +12,37 @@ use std::{str, vec};
 
 /// The location of an import (`use` item) currently being resolved.
 #[derive(PartialEq, Eq)]
-pub struct PendingImport<'fp> {
+struct PendingImport<'fp> {
     filepath: &'fp Path,
     range: ByteRange,
 }
 
 /// A stack of imports (`use` items) currently being resolved.
-pub type PendingImports<'stack, 'fp> = StackLinkedListNode<'stack, PendingImport<'fp>>;
+type PendingImports<'stack, 'fp> = StackLinkedListNode<'stack, PendingImport<'fp>>;
+
+/// Import information(pending imports, glob, and etc.)
+pub struct ImportInfo<'stack, 'fp: 'stack> {
+    imports: PendingImports<'stack, 'fp>,
+    skip_glob: bool,
+}
+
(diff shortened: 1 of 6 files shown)


kngwyu:
I decided to limit the recursion depth via glob to 4.
Before:
```
test glob_imports5 ... bench:  53,956,377 ns/iter (+/- 2,045,050)
test glob_imports6 ... bench: 307,321,133 ns/iter (+/- 34,592,106)
test glob_imports7 ... bench: 2,046,519,408 ns/iter (+/- 161,600,731)
```
After
```
test glob_imports5 ... bench:  44,205,820 ns/iter (+/- 2,150,180)
test glob_imports6 ... bench: 107,402,297 ns/iter (+/- 3,073,601)
test glob_imports7 ... bench: 217,409,533 ns/iter (+/- 5,366,173)
```

It's still slow(maybe we should cache the result of `resolve_path`), but much better.
However, this PR means we give up trying to complete the case where we have to go through the blob imports more than 5 times, e.g.
```
    use mod1::*; use mod2::*; use mod3::*; use mod4::*;
    use Bar::*;
    mod mod1 { pub mod mod2 { pub mod mod3 { pub mod4 {
        pub enum Bar { MyVariant, MyVariant2 }
    }}}}
    MyVa~riant
```

src/racer/matchers.rs
```diff
@@ -12,13 +12,28 @@ use std::{str, vec};
...
 }
 
 /// A stack of imports (`use` items) currently being resolved.
-pub type PendingImports<'stack, 'fp> = StackLinkedListNode<'stack, PendingImport<'fp>>;
+type PendingImports<'stack, 'fp> = StackLinkedListNode<'stack, PendingImport<'fp>>;
+
+/// Import information(pending imports, glob, and etc.)
+pub struct ImportInfo<'stack, 'fp: 'stack> {
+    imports: PendingImports<'stack, 'fp>,
+    glob_depth: Option<usize>,
```
TedDriggs: Can you add a doc comment to this please?

kngwyu:
Thanks for review, done.

kngwyu:
Changed max depth to 3 from 4 in https://github.com/racer-rust/racer/commit/db927a7e946554ea3c57f02f8d28d453f465c054

PR #876, opened 2018-07-04
Bump to 2.1

kngwyu:
We'll be ready to release soon:)

PR #887, opened 2018-07-10
Pass frozen=true flag to cargo by default to avoid network access

kngwyu:
To prevent internet access.
Though sometimes it may be convenient if racer can download dependencies, I think racer should lightweight tool for now.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -54,7 +54,7 @@ build: false
 # TODO modify this phase as you see fit
 test_script:
   - cargo build --verbose
-  - cargo test
+  - cargo test --features allow_net
 
 cache:
   - target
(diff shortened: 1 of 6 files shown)


TedDriggs:
Can you please provide a better writeup of when racer would have gone to the network previously? I'm worried this will increase completion failure rates without people knowing it.

kngwyu:
> Can you please provide a better writeup of when racer would have gone to the network previously?

Sure. Currently, racer downloads a dependency when **there's no specific version of package in local registry**.
So, basically if one add a dependency to `Cargo.toml` and not rebuild it yet, racer downloads it.
It may be convenient, but it can cause significant delay of completion when one's network accesses is unstable.

> I'm worried this will increase completion failure rates without people knowing it.

Yeah, but since the #855 version of racer is not yet released, we don't have to be too nervous about it.

kngwyu:
I wrote some notes in #876

kngwyu:
I found this PR make impossible to get completion via crates-io.patch.... :cold_sweat:

PR #902, opened 2018-07-21
Support macros in stdlibs

kngwyu:
By this PR we can get completions for `prinltn!` or other macros in stdlib.
This implementation is a bit hacky, ~~but since it's difficult to support macros in external crates(to do this, we have to parse too many files :sweat: ), I think it's enough.~~
(Edited: Now I feel we can use this approach for other macros, but not sure)
And, I want to let this the last PR before 2.1.


kngwyu:
I added commits for fixing #903

PR #904, opened 2018-07-22
Get some builtin macros

kngwyu:
#902 lacks support for some builtin macros like `stringify!`, so added it


kngwyu:
Also add support for macros in `libcore/macros.rs` and `liballoc/macros.rs`

PR #910, opened 2018-07-27
Pluggable project model

matklad:
This is a WIP at abstracting the project model, as proposed in https://github.com/racer-rust/racer/pull/855#issuecomment-392448281.

I hope that it'll solve https://github.com/rust-lang-nursery/rls/issues/956, but I haven't implemented the RLS side of things yet, so there's a chance that this whole approach breaks down.

Nevertheless, I figured I'd just toss it over the fence for now to see what the folks think :)

### DIFF
--- a/src/racer/project_model.rs
+++ b/src/racer/project_model.rs
@@ -1,0 +1,221 @@
+use std::path::{Path, PathBuf};
+
+pub trait ProjectModelProvider {
+    fn discover_project_manifest(&self, path: &Path) -> Option<PathBuf>;
+    fn resolve_dependency(&self, manifest: &Path, dep_name: &str) -> Option<PathBuf>;
+}
+
+#[cfg(feature = "cargo")]
+pub(crate) mod cargo {
+    extern crate cargo;
+
+    use std::{
+        collections::{
+            HashSet,
+            hash_map::{self, HashMap},
+        },
+        path::{Path, PathBuf},
+        cell::RefCell,
+        rc::Rc,
+    };
+    use {FileCache, ProjectModelProvider};
+
+    use self::cargo::{
+        Config,
+        ops,
+        util::{errors::CargoResult, important_paths::find_root_manifest_for_wd, toml},
+        core::{
+            Workspace, PackageSet, PackageId, registry::PackageRegistry,
+            resolver::{EncodableResolve, Method, Resolve},
+        }
+    };
+
+    pub fn cargo_project_model<'c>(file_cache: &'c FileCache) -> Box<ProjectModelProvider + 'c> {
+        Box::new(CargoProjectModel {
+            file_cache,
+            cached_lockfile: Default::default(),
+            cached_deps: Default::default(),
+        })
+    }
+
+    struct CargoProjectModel<'c> {
+        file_cache: &'c FileCache,
+        /// Cached lockfiles (path to Cargo.lock -> Resolve)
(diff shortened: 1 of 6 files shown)


kngwyu:
Great!
I feel we need to implement this kind of feature too, but now I don't have enough time, so really thank you for this PR.

But,  I think it's better to merge this PR after implementing `RacerProjectModelProvider` to RLS.

src/racer/project_model.rs
```diff
@@ -0,0 +1,216 @@
+use std::path::{Path, PathBuf};
+
+pub trait RacerProjectModelProvider {
```
kngwyu: Do we really need `Racer` prefix here?
From other crates it looks `racer::RacerProjectModelProvider`

src/racer/project_model.rs
```diff
@@ -0,0 +1,216 @@
...
+    impl<'c> RacerProjectModelProvider for CargoProjectModel<'c> {
+        fn discover_project_manifest(&self, path: &Path) -> Option<PathBuf> {
+            let path = cargo_try!(find_root_manifest_for_wd(path));
+            Some(path)
+        }
+        fn resolve_dependency(&self, manifest: &Path, libname: &str) -> Option<PathBuf> {
+            if self.get_deps(manifest).is_none() {
+                // cache doesn't exist
+                // calculating depedencies can be bottleneck we use info! here(kngwyu)
+                info!("[get_outer_crates] cache doesn't exist");
```
kngwyu: Could you please change function name here?

src/racer/project_model.rs
```diff
@@ -0,0 +1,216 @@
...
+                let deps_map = self.resolve_dependencies(&manifest)?;
+                self.cache_deps(manifest, deps_map);
+            }
+            let deps = self.get_deps(manifest).unwrap();
+            debug!("[get_outer_crates] cache exists for manifest",);
```
kngwyu: Same as above, please change function name in debug!.

src/racer/project_model.rs
```diff
@@ -0,0 +1,216 @@
...
+        }
+        fn resolve_dependency(&self, manifest: &Path, libname: &str) -> Option<PathBuf> {
+            if self.get_deps(manifest).is_none() {
+                // cache doesn't exist
+                // calculating depedencies can be bottleneck we use info! here(kngwyu)
+                info!("[get_outer_crates] cache doesn't exist");
+                let deps_map = self.resolve_dependencies(&manifest)?;
+                self.cache_deps(manifest, deps_map);
+            }
+            let deps = self.get_deps(manifest).unwrap();
```
kngwyu: Why call `self.get_deps` twice here

kngwyu:
I left some comments, but for the whole code, it looks good to me.

matklad:
The RLS sideof things is https://github.com/rust-lang-nursery/rls/issues/956. It's super ugly atm, but at least it works! Will look into clean-up of the racer implementation next! Thanks for the review.

matklad:
Rebased and cleaned-up. [The RLS](https://github.com/rust-lang-nursery/rls/pull/965) side is super incomplete and ugly, but it works, and indeed get's rid of one Cargo :-)

kngwyu:
Now looks good enough to merge :+1: 
But I'm going to wait a little, in case you come up with an alternative idea while working on RLS side :slightly_smiling_face:

PR #911, opened 2018-07-27
Update rustc-ap-syntax to 209.0.0.

kennytm:
The `rustc-ap-syntax` before 209.0.0 will no longer build on future nightly due to rust-lang/rust#51587. Since `rls` depends on `racer 2.1.1`, this PR is needed to unblock the regular RLS builds.


kngwyu:
Thanks!

kennytm:
@kngwyu Could you bump a version and publish to crates.io so that rust-lang/rust#52755 could use it? Many thanks!

kngwyu:
Done

PR #915, opened 2018-08-02
Update rustc-ap-syntax to 211

alexcrichton:
Match the version used by rustfmt (transitively through the RLS) to help ease
build times on rust-lang/rust by only building one copy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -737,7 +737,7 @@ dependencies = [
  "log 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "racer-testutils 0.1.0",
  "rls-span 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


mati865:
Tests failed:
```
failures:
---- nameres::test_do_file_search_std stdout ----
thread 'nameres::test_do_file_search_std' panicked at 'assertion failed: matches.len() > 1', src/racer/nameres.rs:665:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
failures:
    nameres::test_do_file_search_std
test result: FAILED. 74 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

kngwyu:
This test dependes on directry structure of std libs so we don't take care of it here. I'll fit it later.
Thanks!

kngwyu:
This test depends on the directory structure of std lib so we don't take care of it here. I'll fix it later.
Thanks!

PR #918, opened 2018-08-12
Fix overflowing const/static matches (Fixes #746)

### LINKED ISSUE #746: Racer outputs multiple-line results
* Racer version: master. Also present in release 2.0.8

When completing a `const` that occurs any number of lines directly above another `const`, racer matches on not only the identifier of the const, but everything until the end of the line the second const is on as well.

# Reproducing the bug

Given a small file `example.rs`:

```rust
pub const SE



pub const A_CONST: u8 = 2
```

Get completions for the first line:

```bash
racer complete 1 12 example.rs
```

The output is:

```
PREFIX 10,12,SE
MATCH SE



pub const A_CONST,1,10,example.rs,Const,pub const SE
END
```

&nbsp;

&nbsp;

Interestingly, `Debug` output of the `Message` enum containing the match was

```
 Match("SE\n\n\n\npub const A_CONST", 1, 10, "example.rs", Const, "pub const SE")
```

So the problem seems to be caused in the completion machinery itself, which is why https://github.com/racer-rust/racer/pull/585 didn't fix it.

Diomendius:
This fixes behaviour where the match string returned for ``static`` and ``pub const`` (regular ``const`` would trigger this too, but ``is_const_fn`` is broken and stops this as a side-effect) variables and functions will include all characters after the variable name up to the first occurrence of ``':'``. This can be triggered by const declarations with whitespace before the ``':'`` and by certain function declarations such as ``std::Vec::new()``.

Fixes #746

### DIFF
--- a/src/racer/matchers.rs
+++ b/src/racer/matchers.rs
@@ -174,20 +174,22 @@ fn match_pattern_start(
 
     let blob = &src[context.range.to_range()];
     if let Some(start) = find_keyword(blob, pattern, ignore, context) {
-        if let Some(end) = blob[start.0..].find(':') {
-            let s = blob[start.0..start.0 + end].trim_right();
-            return Some(Match {
-                matchstr: s.to_owned(),
-                filepath: context.filepath.to_path_buf(),
-                point: context.range.start + start,
-                coords: None,
-                local: context.is_local,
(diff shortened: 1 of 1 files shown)


kngwyu:
Thanks!

Issue #927, opened 2018-08-18
Openssl dependency build fail


k3yavi:
[log.txt](https://github.com/racer-rust/racer/files/2299582/log.txt)
I am getting the above attached error wile installing racer on Ubuntu using the following command.
```
cargo +nightly install racer
```

Any help resolving it is appreciated .

kngwyu:
How about `sudo apt install libssl-dev`?

kngwyu:
Then.. hmm...
how about using same env var as https://github.com/alexcrichton/ssh2-rs/issues/91#issuecomment-410432033 ?

k3yavi:
Sorry didn't changed a think.
But according to the above issue program is confusing with multiple libraries, i.e. previously installed and one included in ssh2-rs. 
if I am right, then racer don't explicitly install a separate version instead rely on it as an available dependency, isn't it ?

kngwyu:
For some c libraries (e.g. libssh2) racer installs explicit version but for some c libraries(e.g. openssl) it uses system installed one.
But, they're all dependencies of cargo and racer actually didn't need them.
I'm considering remove those dependencies...

k3yavi:
Sure. I am not sure how much work it is to remove the dependencies. But please let me know once you figure it out. Thanks again for the fast replies.

k3yavi:
As a workaround I was able to install the version using stable rust before racer started using nightly builds:

```
 cargo install --version 2.0.14 racer
```

k3yavi:
it worked @kngwyu , thanks for fast resolution ...

Issue #919, opened 2018-08-14
__imp___acrt_iob_func linking errors under msys2


Jimmio92:
Hello,

I am trying to build racer from master branch (pulled yesterday) and it seems to be linking both the GNU C library like it should as well as MSVCRT.. and I'm not sure if that's where the problem lies. Basically, it seems this error occurs when a library is built with MSVC tools prior to VS2015 and then used in VS2015.. It also seems to pop up only in a few curl functions. Any ideas?

kngwyu:
Hmm... It's upstream issue so I'm not sure I can handle it :sweat: 
Now I updated cargo so could you please try to build once more?
But, in the long run, it seems to be better to use `cargo metadata` as subprocess :slightly_frowning_face:

Jimmio92:
Full build log: [buildlog.txt](https://github.com/racer-rust/racer/files/2291667/buildlog.txt)

Nightly Rust. Master branch here. Msys2 up to date. GCC 8.2.0. Windows 10 x64 in a unix-like bash environment, basically.

crlf0710:
@Jimmio92 This has been the case for quite a while now. It's because MSYS2 ships GCC 8.x, and Rust itself ships GCC 7.x. I have some experience dealing with this.

So basically you have two ways to workaround this issue.
1. Somehow get yourself an environment the same with those mingw-build files Rust ships, e.g. GCC 7.x etc.. (Hard)

2. a. remove rust-mingw component. (edit multirust-channel-manifest.toml to change rust-mingw from component to extensions, then rustup component remove rust-mingw).
    b. replace crt2.o and crt2dll.o with the ones from MINGW64/MINGW32 of MSYS2.
    c. build your crate.

This works until you do a `rustup update` then rust-mingw will be back, and you'll have to start over.

Jimmio92:
crt2.o and crt2dll.o are no longer the name of the now multiple files named crtbegin and crtend, plus some mathematical section of the runtime. So.. I don't know what to do. I guess building Rust from source in this environment would alleviate this problem?

crlf0710:
@Jimmio92 those files still exists. there was a typo however, it's called <del>crt2dll.o</del> **dllcrt2.o** .   crtbegin and crtend files do exists, but they're unrelated.

They're located at(taking 64-bit example on my machine):
```
/mingw64/x86_64-w64-mingw32/lib/crt2.o
/mingw64/x86_64-w64-mingw32/lib/dllcrt2.o
```
[...]

PR #930, opened 2018-08-24
migrate to cargo-metadata

kngwyu:
To remove cargo dependency.


kngwyu:
benchmark result
before:
```
test completes_rand  ... bench:  99,061,180 ns/iter (+/- 5,382,882)
test completes_rayon ... bench:  14,012,226 ns/iter (+/- 2,249,149)
```
after: 
```
test completes_rand  ... bench:  66,768,857 ns/iter (+/- 3,299,046)
test completes_rayon ... bench:  29,467,723 ns/iter (+/- 3,618,528)
```

Since I couldn't see a serious performance issue here, I decided to merge this PR.

birkenfeld:
A 100% worse number on `rayon` is not serious?

kngwyu:
@birkenfeld 
Now we have a lot of issues around installing because of cargo(#927, #919), and it's more serious.
Indeed using subprocess make the program slower, but I think 0.01s delay is acceptable.
In addition, since we call `cargo metadata` only once in a session, it shouldn't be a large bottleneck.

birkenfeld:
Well, for many users each completion is a new session. But as long as it's a constant offset that should be ok, at least as long as it doesn't dominate the processing time.

PR #943, opened 2018-09-01
Fix completion for 'for x in ..'

kngwyu:
Enable completion for cases like
```rust
let v: Vec<String> = ..;
for x in v {
    x.~
}
```
.
Current implementation(see https://github.com/racer-rust/racer/blob/1a38903e2640d0cac735f918a858c3596f0222dc/src/racer/typeinf.rs#L250
)
is very inefficient and can't work without manually written `IntoIterator` implementation(i.e. it's useless in 99% cases).


kngwyu:
I ignore some tests now.
First we have to solve #944.

PR #952, opened 2018-09-07
if-let completion

kngwyu:
Enables completion for 
```rust
let s: Option<String> = None;
if let Some(s) = s{
    s.capa~
}
```

Though racer has this feature now(https://github.com/racer-rust/racer/blob/f0e100598b66fd1463bd6c67acf4d90d2806e1b5/src/racer/typeinf.rs#L244), it rarely works.

- [x] less type annotation(e.g. for `let s = Some(String::new());`)


kngwyu:
Now we can get completions for
```rust
let s = Some((String::new(), 0u32));
if let Some((s, _)) = s{
    s.~
}
```

Issue #869, opened 2018-06-13
replace vec::IntoIter with Vec or impl IntoIterator


kngwyu:
Really I don't know why `vec::IntoIter` is used many times.
I think, when we want to evaluate something lazily `impl Iterator` is better, and other cases, just to use `Vec` is better.

h-michael:
@kngwyu 
This issue is now OK to close.

kngwyu:
Closed via #961

PR #961, opened 2018-09-29
 Do not convert iterator until it actually treats iterator

h-michael:
do #869 issue.

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -448,7 +448,8 @@ fn resolve_ast_path(
         core::SearchType::ExactMatch,
         core::Namespace::Path,
         session,
-    ).into_iter()
+    )
+    .into_iter()
     .nth(0)
 }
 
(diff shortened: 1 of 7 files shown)


kngwyu:
Thanks!

birkenfeld:
Not sure this was a good change. It added more LOC, and the API should actually logically change in the other direction (return some `impl Iterator` or `Box<Iterator>`).

(Also, `vec::Vec` is just `Vec`, please.)

kngwyu:
> Also, vec::Vec is just Vec, please.

I'll fix it later.

> return some impl Iterator or Box<Iterator>

I don't think such abstraction is useful to just return multiple matches, since concatenating iterators to Vec is slower than `Vec::append`.

birkenfeld:
> I don't think such abstraction is useful to just return multiple matches, since concatenating iterators to Vec is slower than Vec::append.

Sure. The thing I remember is just that sometimes we only need "the first match"; using iterators would ensure that we don't do unused work. But of course not with the "create a Vec and then call `into_iter` before returning" strategy that was used before this change :)

PR #963, opened 2018-09-29
Fix warning on build

h-michael:
There 4 dead codes.
But I thought that these are intentionally left only because they are not used now, so I exclude them from warning.
Please let me know if this is just a missing.

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -583,14 +583,6 @@ fn myfn(b:usize) {
 }
 
 impl<'c> Src<'c> {
-    pub fn start(&self) -> BytePos {
-        self.range.start
-    }
-
-    pub fn end(&self) -> BytePos {
-        self.range.end
-    }
-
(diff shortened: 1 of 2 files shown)


src/racer/core.rs
```diff
@@ -583,10 +583,12 @@ fn myfn(b:usize) {
 }
 
 impl<'c> Src<'c> {
+    #[allow(dead_code)]
```
kngwyu: These 2 are OK to remove.

src/racer/primitive.rs
```diff
@@ -78,6 +78,7 @@ impl PrimKind {
             LitIntType::Unsuffixed => PrimKind::U32,
         }
     }
+    #[allow(dead_code)]
```
kngwyu: Please do **nothing** for these 2 functions.
They're left for integer method completion but I still haven't decided to implement it for 2.2 or not.

h-michael:
@kngwyu Thanks, I fixed.

kngwyu:
Thanks!

PR #964, opened 2018-09-30
Fix vec::Vec to Vec

h-michael:
Fix this https://github.com/racer-rust/racer/pull/961/files#diff-9b67b9113cab3ef74c5b678891f0bb95R30 .

Related with this comment https://github.com/racer-rust/racer/pull/961#issuecomment-425668456 .


kngwyu:
Thanks!

PR #967, opened 2018-10-04
Suggest to use the official vscode extension.

jens1o:


kngwyu:
I don't know much about `vscode-rust`.
Are you sure we should remove it?

jens1o:
Yeah, because it is no longer maintained. On the other hand, the underlying service of vscode-rust (rls) is updated constantly, and the vscode extension itself also gets some updates, if necessary.

kngwyu:
Thanks, I decided to merge your PR.

jens1o:
Thank you very much!

PR #969, opened 2018-10-10
wip: search for searchstr in body for closure args matches 

### LINKED ISSUE #926: Find definition in use statement can jump into closure arg
Awesome bug (⦿＿⦿)
```rust
use reqwest;
fn f() {
    let a = Some(5).map(|reqwest| {});
}
```
When we use `definition` on `use req~`, it returns  a closure arg below.
It's terrible :fearful:

letmutx:
Hi @kngwyu I am attempting to fix #926 and have a bunch of questions. I am unable to grok the purpose of `point`/`pos`. I thought it is an offset into src, but that is not always the case, which makes sense because parent scopes may not always contain the search term. Can you clear up a bit on how the algorithm for `resolve_path` operates? Thanks.

### DIFF
--- a/src/racer/util.rs
+++ b/src/racer/util.rs
@@ -140,6 +140,7 @@ fn test_find_closure() {
     let src6 = "| x: i32 | Struct { x };";
     let src7 = "y.map(| x: i32 | y.map(|z| z) )";
     let src8 = "|z| z)";
+    let src9 = "let p = |z| something() + 5;";
     let get_range = |a, b| ByteRange::new(BytePos(a as usize), BytePos(b as usize));
     let find = |src: &str, a, off1: i32, b, off2: i32| {
         get_range(src.find(a).unwrap() as i32 + off1, src.rfind(b).unwrap() as i32 + 1 + off2)
@@ -152,8 +153,9 @@ fn test_find_closure() {
     assert_eq!(Some((get_pipe(src4), find(src4, 's', 0, ')', 0))), find_closure(src4));
     assert_eq!(Some((find(src5, '|', 0, 'y', -2), find(src5, 'y', 0, ')', 0))), find_closure(src5));
     assert_eq!(Some((get_pipe(src6), find(src6, 'S', 0, ';', -1))), find_closure(src6));
-    assert_eq!(Some((find(src7, '|', 0, 'y', -2), find(src7, '2', 4, 'z', 1))), find_closure(src7));
-    assert_eq!(Some((get_pipe(src8), find(src8, ' ', 1, 'z', 0))), find_closure(src8));
+    assert_eq!(Some((find(src7, '|', 0, 'y', -2), find(src7, '2', 4, ')', 0))), find_closure(src7));
+    assert_eq!(Some((get_pipe(src8), find(src8, ' ', 1, ')', 0))), find_closure(src8));
+    assert_eq!(Some((get_pipe(src9), find(src9, 's', 0, '5', 0))), find_closure(src9));
 }
 
 /// Try to valid if the given scope contains a valid closure arg scope.
--- a/src/racer/util.rs
+++ b/src/racer/util.rs
@@ -79,6 +79,83 @@ pub fn symbol_matches(stype: SearchType, searchstr: &str, candidate: &str) -> bo
     }
 }
 
+pub fn find_closure(src: &str) -> Option<(ByteRange, ByteRange)> {
+    let (pipe_range, _) = closure_valid_arg_scope(src)?;
+    let mut chars = src.chars()
+                       .enumerate()
+                       .skip(pipe_range.end.0)
+                       .skip_while(|(_, c)| {
+                           c.is_whitespace()
+                       });
+    let (start, start_char) = chars.next().map(|(i, c)| (if c == '{' { i + 1 } else { i }, c))?;
+
+    let mut clevel = if start_char == '{' { 1 } else { 0 };
+    let mut plevel = 0;
+
+    let mut last = None;
+    for (i, current) in chars {
+        match current {
+            '{' => clevel += 1,
+            '(' => plevel += 1,
+            '}' => {
+                clevel -= 1;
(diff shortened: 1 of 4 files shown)


kngwyu:
>  I am unable to grok the purpose of point/pos

Please link the point in code, and please question in #926, not here.

This PR doesn't look **bug fix** of #926, but of a temporal hack.
So I don't want to this PR now, sorry.
Since historical reasons, racer's code is ugly and not readable(I'm not a original author as you, so I often it's hard too understand, too).
So I highly recommend you to tackle easier bugs first, like [these](https://github.com/racer-rust/racer/issues?q=is%3Aissue+is%3Aopen+label%3Agood-first-issue).

kngwyu:
How about modifying `closure_valid_arg_scope`, instead of adding `find_closure_body`?

> I am unable to grok the purpose of point/pos
Could you please write the question in #926, with clarifying where `point`/`pos` you mean by source code link?
Sorry.

kngwyu:
In addition, I want at least one test in tests dir.

src/racer/util.rs
```diff
@@ -79,6 +79,87 @@ pub fn symbol_matches(stype: SearchType, searchstr: &str, candidate: &str) -> bo
...
+    };
+    let mut chars = chars.peekable();
+    while chars.peek().is_some() {
+        let (i, current) = chars.next().unwrap();
+        match current {
+            '{' => clevel += 1,
+            '(' => plevel += 1,
+            '}' => {
+                clevel -= 1;
+                if clevel == 0 && (start_char == '{') {
```
kngwyu: `clevel == 0 && start_char == '{'`

src/racer/util.rs
```diff
@@ -79,6 +79,87 @@ pub fn symbol_matches(stype: SearchType, searchstr: &str, candidate: &str) -> bo
...
 
+pub fn find_closure(src: &str) -> Option<(ByteRange, ByteRange)> {
+    let (pipe_range, _) = closure_valid_arg_scope(src)?;
+    let mut chars = src.chars()
+                       .enumerate()
+                       .skip(pipe_range.end.0)
+                       .skip_while(|(_, c)| {
+                           c.is_whitespace()
+                       });
+    let (start, start_char) = if let Some((i, c)) = chars.next() {
```
kngwyu: `chars.next().map(|(i, c)| (if c == '{' { i ; 1} else {i}, c))?;`

src/racer/util.rs
```diff
@@ -79,6 +79,87 @@ pub fn symbol_matches(stype: SearchType, searchstr: &str, candidate: &str) -> bo
...
+        (if c == '{' { i + 1 } else { i }, c)
+    } else {
+        return None;
+    };
+    let (mut clevel, mut plevel) = if start_char == '{' {
```
kngwyu: I feel
```rust
let mut clevel = if start_char == '{' { 1 } else { 0 };
let mut plevel = 0;
```
is better, but you don't have to modify if you don't like it.

src/racer/util.rs
```diff
@@ -79,6 +79,87 @@ pub fn symbol_matches(stype: SearchType, searchstr: &str, candidate: &str) -> bo
...
+                }
+            },
+            ')' => {
+                plevel -= 1;
+                if plevel == -1 {
+                    return Some((pipe_range, ByteRange::new(BytePos(start), BytePos(i))));
+                }
+                if plevel == 0 && start_char != '{' {
+                    match chars.peek() {
+                        Some((_, c)) if *c == '.' => {}
```
kngwyu: I don't like this heuristics, because it has many edge cases like `[`.
How about just waiting `;`?

kngwyu: I don't like this heuristics since it has many edge cases(e.g. `a.method()[] or a.method() + 5` is also can be available.)
How about just waiting `;` or `plevel == -1`

letmutx:
`rustfmt` was adding formatting changes which are not related to code changes, I wasn't sure if that's okay. May be you should to run `cargo fmt` on master and push all the formatting changes.

Using `libsyntax` didn't even cross my mind. It would really simplify this. I will update the PR accordingly. Thanks for the review!

letmutx:
I tried using `libsyntax` for parsing the closure but it was failing when parsing incomplete bodies. So, did a greedy approach for parsing closure body. Please let me know if you need any changes!

src/racer/util.rs
```diff
@@ -79,6 +79,88 @@ pub fn symbol_matches(stype: SearchType, searchstr: &str, candidate: &str) -> bo
...
+            ';' => {
+                if start_char != '{' {
+                    last = Some(i);
+                    break;
+                }
+            },
+            ')' => {
+                plevel -= 1;
+                if plevel == 0 {
+                    last = Some(i + 1);
```
kngwyu: Is this necessary?

src/racer/util.rs
```diff
@@ -79,6 +79,88 @@ pub fn symbol_matches(stype: SearchType, searchstr: &str, candidate: &str) -> bo
...
+                           c.is_whitespace()
+                       });
+    let (start, start_char) = chars.next().map(|(i, c)| (if c == '{' { i + 1 } else { i }, c))?;
+
+    let mut clevel = if start_char == '{' { 1 } else { 0 };
+    let mut plevel = 0;
+
+    let mut chars = chars.peekable();
+    let mut last = None;
+    while chars.peek().is_some() {
```
kngwyu: Now looks just `for (i, current) in chars` is enough.

kngwyu:
@letmutx 
> it was failing when parsing incomplete bodies

Thanks for tackling a hard problem!
I think it's really helpful for future implementation.

I left some comments about refactoring, so let's merge this PR after resolving them.

src/racer/util.rs
letmutx: MistressMissed itit, sorry!

letmutx: Missed it, sorry!

PR #970, opened 2018-10-11
Bump rustc-ap-syntax to 272.0.0

otavio:
This bumps the dependency to the same we are updating in `rustfmt` so
`rustc` can use a single version globally.

Refs: https://github.com/rust-lang-nursery/rustfmt/pull/3086

Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>


kngwyu:
Thanks!

PR #971, opened 2018-10-12
Add closure completions

letmutx:
Adds support for closure return types; There is no functionality for closures which are not annotated ([link](https://github.com/racer-rust/racer/blob/master/src/racer/ast.rs#L1273)) so I haven't spent time on supporting cases like:
```rust
fn main() {
    let f = |x: i32| Some(x * 2);
    f().un~
}
```
Thanks!

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -636,6 +636,19 @@ impl<'c, 's, 'ast> visit::Visitor<'ast> for ExprTypeVisitor<'c, 's> {
                                 }
                                 Some(Ty::Match(m))
                             }
+                            MatchType::TypeParameter(ref traitbounds) if traitbounds.has_closure() => {
+                                let mut output = None;
+                                if let Some(path_search) = traitbounds.get_closure()
+                                {
+                                    for seg in path_search.path.segments.iter() {
+                                        if seg.output.is_some() {
+                                            output = seg.output.clone();
+                                            break;
+                                        }
+                                    }
+                                }
+                                output
+                            }
                             _ => {
                                 debug!(
                                     "ExprTypeVisitor: Cannot handle ExprCall of {:?} type",
--- a/src/racer/ast_types.rs
+++ b/src/racer/ast_types.rs
@@ -64,6 +64,7 @@ impl Ty {
         }
         ty.wrap_by_ref(deref_cnt)
     }
+
     pub(crate) fn dereference(self) -> Self {
         if let Ty::RefPtr(ty, _) = self {
             ty.dereference()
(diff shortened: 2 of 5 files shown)


letmutx:
Please let me know if there are any issues! Failing tests are not related, as far as I understand

src/racer/ast.rs
```diff
@@ -636,6 +636,21 @@ impl<'c, 's, 'ast> visit::Visitor<'ast> for ExprTypeVisitor<'c, 's> {
                                 }
                                 Some(Ty::Match(m))
                             }
+                            MatchType::TypeParameter(ref traitbounds) if traitbounds.has_closure() => {
+                                let mut output = None;
+                                if let Some(path_search) = traitbounds.find_by_name("Fn")
```
kngwyu: How about adding `TraitBounds::find_by_name(&self, names: &[&str]) -> Option<&PathSearch>`?

src/racer/ast_types.rs
```diff
@@ -449,11 +470,13 @@ impl fmt::Display for Path {
 pub struct PathSegment {
     pub name: String,
     pub generics: Vec<Ty>,
+    /// If this path segment is a closure, it's return type
+    pub output: Option<Ty>,
```
kngwyu: I think 
```
generics: Option<GenericsKind>,
...
enum GenericsKind {
   Angle(Vec<Ty>),
   Paren{in: Vec<Ty>, out: Option<Ty> },
}
```
is better.
But, since it needs large refactoring, you don't have to fix it in this PR.
I'll fix it later.

src/racer/ast_types.rs
```diff
@@ -695,7 +763,50 @@ impl GenericsArgs {
                 _ => {}
             }
         }
-        GenericsArgs(args)
+
+        // resolve the closure's return type into the containing function's type parameter
+        fn replace_closure_output_with_matching_type_params_from_fn(tp: &mut TypeParameter, args: &GenericsArgs) {
+            if let Some(ps) = tp.bounds.get_closure_mut() {
+                for segment in ps.path.segments.iter_mut() {
+                    let s = match segment.output.take() {
```
kngwyu: Maybe `and_then` is better.

src/racer/ast_types.rs
```diff
@@ -695,7 +763,50 @@ impl GenericsArgs {
                 _ => {}
             }
         }
-        GenericsArgs(args)
+
+        // resolve the closure's return type into the containing function's type parameter
+        fn replace_closure_output_with_matching_type_params_from_fn(tp: &mut TypeParameter, args: &GenericsArgs) {
```
kngwyu: How about returning `Option<()>` here?
Then we can use `?` to shorten code.

kngwyu:
Left some comments, but I feel this is really great enhancement!
Thanks, @letmutx !

src/racer/ast.rs
letmutx: I added a method `get_closure` but forgot to update this location.

kngwyu:
Merged now, thanks again.

PR #976, opened 2018-10-19
Add type inference for binary operators

### LINKED ISSUE #901: Type inference for binary operator
e.g.
```rust
let a = Point(5, 6) - Point(6, 7);
a.~ 
```

Though it doesn't look easy :cold_sweat:

letmutx:
Attempt to fix #901


src/racer/ast.rs
```diff
@@ -841,6 +838,94 @@ impl<'c, 's, 'ast> visit::Visitor<'ast> for ExprTypeVisitor<'c, 's> {
...
+                        Some(m) => m.matchstr,
+                        _ => {
+                            return;
+                        }
+                    },
+                    _ => {
+                        return;
+                    }
+                };
+                let searchstr = match bin.node {
```
kngwyu: I think these codes should go to `typeinf.rs`.
We shouldn't make a function too long.

src/racer/ast.rs
```diff
@@ -841,6 +838,94 @@ impl<'c, 's, 'ast> visit::Visitor<'ast> for ExprTypeVisitor<'c, 's> {
...
+                                };
+                            }
+                            // default is self
+                            return trait_impl.self_path().name() == Some(other_type);
+                        }
+                    }
+                    false
+                };
+
+                let resolve_binary_expr_type = |contextm: &Match, other_type: &str| {
```
kngwyu: Same as the above comment, how about moving this function to `nameres.rs`?

tests/system.rs
```diff
@@ -3907,12 +3907,12 @@ fn follows_multiline_use() {
 
 #[test]
 fn follows_external_re_export() {
-    let src = "
+    let src = r"
```
kngwyu: What is the intention of adding `r`?

tests/system.rs
```diff
@@ -4051,3 +4051,137 @@ fn completes_functions_from_trait_objects() {
     let got = get_all_completions(src, None);
     assert!(got.into_iter().any(|ma| ma.matchstr == "foo"));
 }
+
+#[test]
```
kngwyu: Should be in other file, like `tests/binary.rs`.

kngwyu:
Great, thanks!
But your branch looks based on a bit old master.
Please rebase onto current master, which is formatted by rustfmt :)

tests/system.rs
```diff
@@ -3862,7 +3862,7 @@ fn finds_std_macro_doc() {
     fn main() {
         module_pat~h!();
     }
-    "#;
```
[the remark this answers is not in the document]
kngwyu: I don't care.

letmutx:
Failures don't seem to be related to tests. Is there a way to rerun them?

kngwyu:
@letmutx 
It looks appveyor's bug and we don't have to mind it.

But, because of https://github.com/racer-rust/racer/commit/9207d5e7b3a554397ff7d70121b7facb86ca2849, need to rebase again.
Sorry for taking your time again.

src/racer/nameres.rs
```diff
@@ -2450,8 +2462,57 @@ pub(crate) fn get_index_output(selfm: &Match, session: &Session) -> Option<Ty> {
...
+/// Resolves the type of a binary expression
+/// # Arguments
+/// * base_type: the type on the left hand side
+/// * node: the operator
+/// * other_type: the type on the right hand side
+/// *Note*: Works only for std::ops::* now
+pub(crate) fn resolve_binary_expr_type(
+    base_type: &Match,
+    node: BinOpKind,
+    other_type: &str,
```
kngwyu: This should be `Option<&str>`.
I think we should inference that `L + R = L` if we don't know R's type.

letmutx: Does it make sense to check if there's an override impl to reduce false positives?

kngwyu: I don't think so.
The important point is **we can't always find operator impl**.
(e.g. If we use [derive_more](https://github.com/JelteF/derive_more) and write
```rust
#[derive(Add, Sub)]
struct S(u64);
```, or place impl in other file than struct definition file).
So I think to assume that L=R  is OK if
- we can't find R's type
- we can'T find trait implementation

File: src/racer/nameres.rs [hunk 7]
```diff
@@ -2488,8 +2500,70 @@ pub(crate) fn get_index_output(selfm: &Match, session: &Session) -> Option<Ty> {
...
+        session,
+    )
+    .into_iter()
+    .filter(|trait_impl| has_impl_for_other_type(trait_impl, trait_name, other_type))
+    .next();
+    if let Some(matching_impl) = matching_impl {
+        get_associated_type_match(&matching_impl, "Output", &base_type, session)
+            .or_else(|| Some(Ty::Match(base_type.clone())))
+    } else {
+        // default to base type if an impl can't be found
```
kngwyu: :+1:

tests/binary.rs
```diff
@@ -0,0 +1,136 @@
...
+             Bar
+        }
+    }
+    fn main() {
+        let a = Left { x: 1, y: 0 } - Right { x: 2, y: 3 };
+        a.~
+    }
+";
+    let got = get_one_completion(src, None);
+    assert_eq!(got.matchstr, "foo".to_string());
```
kngwyu: just `"foo"`

tests/binary.rs
```diff
@@ -0,0 +1,136 @@
...
+             Bar
+        }
+    }
+    fn main() {
+        let a = Left { x: 1, y: 0 } - Left { x: 2, y: 3 };
+        a.~
+    }
+";
+    let got = get_one_completion(src, None);
+    assert_eq!(got.matchstr, "bar".to_string());
```
kngwyu: Same as above, no need to use `to_string`

src/racer/nameres.rs
```diff
@@ -2488,8 +2500,70 @@ pub(crate) fn get_index_output(selfm: &Match, session: &Session) -> Option<Ty> {
...
+/// * base_type: the type on the left hand side
+/// * node: the operator
+/// * other_type: the type on the right hand side
+/// *Note*: Works only for std::ops::* now
+pub(crate) fn resolve_binary_expr_type(
+    base_type: &Match,
+    node: BinOpKind,
+    other_type: Option<&str>,
+    session: &Session,
+) -> Option<Ty> {
```
kngwyu: Now this function doesn't fail so should return `Match`

src/racer/nameres.rs
```diff
@@ -2488,8 +2500,70 @@ pub(crate) fn get_index_output(selfm: &Match, session: &Session) -> Option<Ty> {
...
+    trait_impl: &ImplHeader,
+    trait_name: &str,
+    other_type: Option<&str>,
+) -> bool {
+    if let Some(ref path) = trait_impl.trait_path() {
+        if path.name() == Some(trait_name) {
+            if other_type.is_none() && path.segments[0].generics.len() == 0 {
+                return true;
+            }
+            if path.segments[0].generics.len() > 0 {
```
kngwyu: What does it mean?
Is it for cases like `R = St<String>` and `Output = St<T>`?
Since trait_name is just a path name and doesn't include generics, I think it's impossible to judge if the R and Output's type parameter matches.

kngwyu:
I left some comments.
Sorry for taking your time again.

src/racer/nameres.rs
letmutx: But let's keep it option. We don't need to refactor if we enhance in the future

kngwyu: > We don't need to refactor if we enhance in the future

I don't feel it 'enhance' to return `Option` here

letmutx: What if you could always find an impl and there isn't a valid impl for this type? You have to return `None`. That said, I am not against returning just a `Match`

src/racer/nameres.rs
letmutx: No. It is for cases like `R = String` and `impl Add<String>...` I am trying to match `String` here. I should also recursively check generics for types like `R = MyType<String>` and `impl Add<MyType<String>> for ...` which I obviously missed. I will update the PR. Thanks.

src/racer/nameres.rs
kngwyu: > What if you could always find an impl

I think it's very hard... but I also feel that  we shouldn't always returns fallback(base_type) too.

Let's leave it as it is now, and let's change it if we come up with a better idea.

src/racer/nameres.rs
kngwyu: Ah I understand.
But I think `generics[0]` can be other than PathSearch(e.g. `&Point`).
Maybe we need some utility method for type matching?

PR #979, opened 2018-10-21
interner: replace explicit instances for InternedString by derive

birkenfeld:


interner/src/lib.rs
```diff
@@ -65,12 +61,6 @@ impl Deref for InternedString {
     }
 }
 
-impl Hash for InternedString {
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        Hash::hash(&(self.inner as *const str), state)
```
kngwyu: This is different from `Derive` d implementation and I want to leave it

kngwyu:
Thanks!
I left a comment but other than that looks good.

interner/src/lib.rs
birkenfeld: Ah, indeed! I didn't realize that raw pointers do not hash transparently.

Why is it necessary to not hash the same as a non-interned string?

kngwyu: Just for speed up.
But, since I replaced most of `HashMap<InternedString, _>` with `Vec`, now we don't have to be too nervous about it.
To use `derive` is also OK if you preffer to it.

birkenfeld: If there is a good reason, I have no objections (other than that it should probably be commented...)

kngwyu: I see

kngwyu: I see

Issue #916, opened 2018-08-05
find-definition requires `extern crate` in 2018 edition


tarka:
`extern crate` is no longer required in the 2018 edition preview. However, trying `find-definition 3 25` on the following code fails. However if add `extern crate actix_web;` it works.

    #![feature(rust_2018_preview)]

    use actix_web::{server, App, HttpRequest};

    fn index(_req: &HttpRequest) -> &'static str {
        "Hello world!"
    }

    fn main() {
       server::new(|| App::new().resource("/", |r| r.f(index)))
            .bind("127.0.0.1:8088")
            .unwrap()
            .run();
    }

(This is built from master 22033f9, using emacs-racer, but confirmed on the commandline.)

kngwyu:
Thanks!
It's just not implemented yet :sweat:

kngwyu:
By #930 racer know current project's edition.
Next step is to know 'where the current crate is'.

tav:
Is anyone planning on working on this by any chance?

kngwyu:
I'm worrying about the overhead of calling `cargo metadata` just to know what the edition is.
But, anyway, I'm going to include this in 2.2 release and any contribution is welcome.

kngwyu:
Closed via #987

PR #987, opened 2018-10-25
 Completes crate name without extern 

kngwyu:
For #916


kngwyu:
Hmm `doesnt_complete_cratename_without_extern_in_2015` test is failed in CI.
It's weird :roll_eyes:

PR #989, opened 2018-10-28
Fix name shadowing

### LINKED ISSUE #809: Local variable shadows `::name`
In the following code:

```rust
fn name() { unimplemented!() }

#[test]
fn test_name() {
    let name = 1;
    let value = ::name();
}
```

Go to definition on `::name` jumps to the local variable `name`, but it should jump to the function outside.

letmutx:
Attempt to fix #809

### DIFF
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -861,13 +861,16 @@ pub fn search_crate_root(
     namespace: Namespace,
     session: &Session,
     import_info: &ImportInfo,
+    skip: bool,
 ) -> Vec<Match> {
     debug!("search_crate_root |{:?}| {:?}", pathseg, modfpath.display());
 
     let crateroots = find_possible_crate_root_modules(modfpath.parent().unwrap(), session);
     let mut out = Vec::new();
     for crateroot in crateroots {
-        if *modfpath == *crateroot {
+        // when searching paths with global prefix, we don't
+        // want to skip checking the current file, but in other cases, we do
+        if *modfpath == *crateroot && skip {
             continue;
         }
         debug!(
@@ -1498,6 +1501,7 @@ pub fn resolve_name(
         namespace,
         session,
         import_info,
+        true,
     ) {
         out.push(m);
         if is_exact_match {
@@ -1768,6 +1772,16 @@ pub fn resolve_path(
                     return Vec::new();
                 }
             }
+            PathPrefix::Global => {
+                return resolve_global_path(
+                    path,
+                    filepath,
+                    search_type,
+                    namespace,
+                    session,
+                    import_info,
+                );
+            }
             _ => {}
         }
     }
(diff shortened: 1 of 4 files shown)


src/racer/nameres.rs
```diff
@@ -861,13 +861,21 @@ pub fn search_crate_root(
     namespace: Namespace,
     session: &Session,
     import_info: &ImportInfo,
+    skip: bool,
```
kngwyu: Please use more concrete name, like `skip_this_file` or `skip_if_self_is_mod`.
But, I don't have a good idea now :slightly_frowning_face: , sorry.

src/racer/nameres.rs
```diff
@@ -1819,6 +1838,54 @@ pub fn resolve_path(
...
+    if let Some(context) = context {
+        if path.segments.len() == 1 {
+            let mut v = Vec::new();
+            v.push(context);
+            return v;
+        }
+        let mut new_path = path.clone();
+        new_path.prefix = None;
+        match context.mtype {
+            MatchType::Module => resolve_path(
```
kngwyu: Here `MatchType::Crate` can be available so please use `resolve_following_path` here.

tests/system.rs
```diff
@@ -3807,3 +3807,52 @@ fn import_stmt_doesnt_jump_to_closure_arg() {
...
+    let src = r#"
+    fn foo() {
+        let a = ::na~me();
+    }
+"#;
+    let lib = r#"
+    use mod1;
+    pub(crate) fn name() {}
+"#;
+    let (pos, src) = get_pos_and_source(src);
```
kngwyu: Please use 
```rust
    let dir = TmpDir::new();
    let _lib = dir.write_file("lib.rs");
    let got = get_definition(src, Some(dir));
```

src/racer/nameres.rs
```diff
@@ -1768,6 +1777,16 @@ pub fn resolve_path(
                     return Vec::new();
                 }
             }
+            PathPrefix::Global => {
```
kngwyu: :+1:

kngwyu:
I left some comments, but other than that looks great.
Thanks!

tests/system.rs
letmutx: I did use that first and the tests were failing; because `search_crate_root` wasn't able to find the root file.

src/racer/nameres.rs
letmutx: Me too :) So added a comment where it is used. Would it be better if we change the `modfpath` to an `Option`?

kngwyu: Since we use `modfpath` to get parent directory we can't do that.
How about `skip_modfpath`?

kngwyu:
Hi, @letmutx, how is the progress of this PR?
Or, if you're busy, please feel free to let me do the rest work.

letmutx:
I'll not be able to take a look until this weekend. Please continue if you get a chance. Thanks.

kngwyu:
Thank for replaying.
Suddenly I have no time too...

PR #1000, opened 2018-11-16
Re-export Edition

nrc:
Currently `ProjectModelProvider` cannot be implemented because there is no way to name `Edition`.


kngwyu:
Thanks, I forgot to modify rls side

PR #1007, opened 2018-12-07
Correctly count CRLF newlines when counting line offsets.

akovaski:
This allows completion coordinates to find the correct position in the file when using Windows style newlines (CRLF). I added two simple tests which are slight tweaks of the coords_to_point_works test in core.rs.

Notably for me, this allows RLS autocomplete to function correctly when working on files with Windows style newlines.

See [issue 976 in the rust-lang/rls repo](https://github.com/rust-lang/rls/issues/976). I believe this behavior was merged to the racer github repo on Jun 24, 2018 with commit 42046fc27cf321dc7c0fc70bdf11e8e4a024b32a.

Example image of the autocomplete problem in VSCode on Windows. This is a file with CRLF newlines. Notice that the cursor is on line 7, but the auto-completion is for the code that exists on line 5.
image

Now the autocomplete will autocomplete line 5 at the end of line 5, where the user will want it to. 🎉 
image

### DIFF
--- a/src/racer/core.rs
+++ b/src/racer/core.rs
@@ -487,7 +487,7 @@ impl RawSource {
         let mut before = 0;
         *self.lines.borrow_mut() = self
             .code
-            .lines()
+            .split('\n')
             .map(|line| {
                 let len = line.len() + 1;
                 let res = ByteRange::new(before, before + len);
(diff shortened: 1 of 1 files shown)


Xanewok:
Nice find! This actually fixes Windows CI failure for RLS, which was mismatched auto-completion based on lines https://travis-ci.org/Xanewok/rls/jobs/465194944 (errors because of debug `Error: StringError("actually ok")` but it's actually fixed now) :tada:

kngwyu:
Thanks :+1: 
@Xanewok 
Should I publish new version for rls?

Xanewok:
@kngwyu I think that’d be great, yes. Windows users get their autocompletion back so I think that alone meets the point release bar 👍

kngwyu:
Published v2.1.15 now

kngwyu:
I yanked 2.1.15 because of really silly miss https://github.com/racer-rust/racer/commit/4c2d363671937eef96e7d28ac103cd8545c90ecd
Please use 2.1.16 :sweat:

Xanewok:
Thanks for the release!

PR #1010, opened 2018-12-23
Fix parsing of documentation block comments inside block comments

Nuramon27:
When a source file contained a documentation block comment `/** doc comment */` that was nested into a block comment, i. e. something like
```
/* this is a /** documentation block comment inside a */ block comment */
```
which might occur when someone uncomments a function together with its documentation, racer failed to provide completions in the entire file.

The problem seems to lie in the method `comment_block` of `codecleaner::CodeIndicesIter`: When reaching the `/**` while parsing the comment, `prev` is first set to `b'/'`. After reaching the first `*`, the nesting level is increased, but `prev` is still `b'/'`, so when the second `*` is reached, the nesting level is increased again, as the previous character is still considered to be a slash.

I was not able to figure out why this does not happen on unnested documentation block comments.

I added a test that tests the behaviour in this case and fixed the `comment_block` by setting `prev` to a whitespace character after `/*` or `*/` is found.

### DIFF
--- a/src/racer/codecleaner.rs
+++ b/src/racer/codecleaner.rs
@@ -116,6 +116,7 @@ impl<'a> CodeIndicesIter<'a> {
             pos = pos.increment();
             match b {
                 b'/' if prev == b'*' => {
+                    prev = b' ';
                     if nesting_level == 0 {
                         break;
                     } else {
@@ -123,6 +124,7 @@ impl<'a> CodeIndicesIter<'a> {
                     }
                 }
                 b'*' if prev == b'/' => {
+                    prev = b' ';
                     nesting_level += 1;
                 }
                 _ => {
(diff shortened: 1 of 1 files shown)


kngwyu:
Thanks!
Looks really reasonable change for me.
But how about 
```
            match b {
                b'/' if prev == b'*' => {
                    if nesting_level == 0 {
                        break;
                    } else {
                        nesting_level -= 1;
                    }
                }
                b'*' if prev == b'/' => {
                    nesting_level += 1;
                }
                _ => (),
            }
            prev = b;
```
?

Nuramon27:
Im afraid this would not be correct because then in
```
/* outer comment block /*/ racer woud consider this to be the outer block */ … and this to be code */
```
already the second slash in `/*/` would be considered as the and of the inner comment.

kngwyu:
Ah, I understand, thanks.

PR #1016, opened 2019-02-11
Update rustc-ap-syntax from 306.0.0 to 366.0.0

h-michael:
rustfmt use v366.0.0.
So if we use same version of rustc-ap-syntax, we can't update rust submodule.
related with: #58356

@kngwyu 
After merge this, would you release new racer version?

### DIFF
--- a/src/racer/ast_types.rs
+++ b/src/racer/ast_types.rs
@@ -144,6 +144,7 @@ impl Ty {
             }
         }
     }
+
     pub(crate) fn from_lit(lit: &ast::Lit) -> Option<Ty> {
         let make_match = |kind: PrimKind| kind.to_module_match().map(Ty::Match);
         match lit.node {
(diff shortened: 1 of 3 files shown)


src/racer/ast_types.rs
```diff
@@ -159,6 +160,8 @@ impl Ty {
             },
             LitKind::FloatUnsuffixed(_) => make_match(PrimKind::F32),
             LitKind::Bool(_) => make_match(PrimKind::Bool),
+            // TODO: support err
+            LitKind::Err(_) => None,
```
h-michael: The new kind added with [this PR](https://github.com/rust-lang/rust/pull/57651).

kngwyu: We don't have no item to complete for bad literal like `'aaa'`, so I don't think we need TODO comment here.

kngwyu:
Thanks :wink: 
I've left just a trivial comment, but otherwise LGTM :+1:

h-michael:
@kngwyu Thank you for your quick reply. I've fixed and push :)

kngwyu:
Thanks again :+1:

kngwyu:
Uploaded 2.1.18 now

h-michael:
@kngwyu Thanks :)

PR #1018, opened 2019-02-13
Bump rustc-ap deps to v373

Xanewok:
Context: https://github.com/rust-lang/rust/pull/58337#issuecomment-463396117
See also https://github.com/rust-lang/rustfmt/pull/3345.

Coordinated update to unbreak current nightlies. If this is merged, please release a new patch version to crates.io with those updates, thanks! (This only compiles on newest Rust master, so we need to wait a couple of hours for newest nightlies).


topecongiro:
FWIW just published rustfmt [1.0.3](https://crates.io/crates/rustfmt-nightly/1.0.3).

kngwyu:
Thanks!

PR #1021, opened 2019-02-20
Sort matches by matchstr and by point before dedup

rpeshkov:
Fix for issue https://github.com/rust-lang/rls-vscode/issues/506

Dedup function was not removing all duplicate entries, because matches vector is not sorted. This PR introduces sorting of matches vector by matchstr first and then by match point before dedup.


kngwyu:
Thanks.

jens1o:
@kngwyu I wonder whether you could file a release with this commit included so https://github.com/rust-lang/rls/issues/287 can get fixed?

PR #1027, opened 2019-03-17
Prefer `debug!` over `println!` in library code

Xanewok:
This should fix completion in RLS since we can't print anything to stdout ourselves (due to LSP), leading to a malformed state.

I'm sorry @kngwyu, could I ask you to release a new version one more time? :bowing_man: I'd like to ship the version with rustc-ap-* v407 but don't want to ship RLS hanging on enum autocompletions :cry:

### DIFF
--- a/src/racer/nameres.rs
+++ b/src/racer/nameres.rs
@@ -1613,7 +1613,7 @@ fn get_enum_variants(
     session: &Session,
 ) -> Vec<Match> {
     let mut out = Vec::new();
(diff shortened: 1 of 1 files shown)


kngwyu:
OMG, sorry maybe I forgot to erase it :sweat:

Xanewok:
Thank you for such a quick response!

kngwyu:
Published 2.1.21 now

PR #1036, opened 2019-04-20
Update rls-span to 0.5.1

Xanewok:
This version brings serde support but with optional derive macros
for (de)serialization (Rust can't use proc macros from crates.io yet), as part of a coordinated update to enable serde support for save-analysis files in the compiler.

r? @kngwyu 

When/if that lands, I'd very much appreciate a patch release with this, if possible - thanks!


kngwyu:
Thanks, I'll publish it soon.

kngwyu:
Published 2.1.22 now

Xanewok:
Awesome, thank you!

PR #1048, opened 2019-06-09
Cargo update

topecongiro:
Update rustc-ap-syntax to 491.0.0.

Note that I am just passing edition 2018 to `syntax::with_globals`. Ideally, we should pass the correct edition to it, but I find it to require quite a lot of code refactoring, and I basically gave up. Since the current implementation of racer does not seem to properly set the correct edition anyway, I am assuming that there will be no breaking change introduced with this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "annotate-snippets"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "ansi_term"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


topecongiro:
cc https://github.com/rust-lang/rustfmt/pull/3611.

kngwyu:
Thanks!

> Since the current implementation of racer does not seem to properly set the correct edition anyway, I am assuming that there will be no breaking change introduced with this PR.

Thanks you for trying that.
The problem is calling `cargo metadata` can be slow but anyway we have to do do that

topecongiro:
Could you make a point release so that we could update the dependencies of rls?

kngwyu:
Sure. Now I'm doing some bug fixes and I'll release it by today.

kngwyu:
Released 2.1.23 now.

topecongiro:
@kngwyu Thank you!

Issue #1067, opened 2019-08-25
src/racer/lib.rs marked as executable


hvenev:
src/racer/lib.rs marked as executable but has a `#!` with a relative path. This confuses some tools and most notably makes Fedora's build system fail to build the rust package.

kngwyu:
Resolved by #1068

PR #1068, opened 2019-08-25
Mark rust source file as non-executable.

hvenev:
Fixes issue #1067

It would be nice if you can backport this to beta as well.


kngwyu:
Thank you!

> It would be nice if you can backport this to beta as well.

What does *beta* mean?
2.0.*?

hvenev:
I mean rustc 1.38.

kngwyu:
Now racer supports only nightly :scream: , sorry.

hvenev:
A version of racer is included in the rustc beta release tarball. Can you include this fix in the version that gets in the stable tarball?

h-michael:
@hvenev To guess, isn't it a racer that goes in with rls dependencies?

PR #1076, opened 2019-10-06
Update rustc-ap-syntax to 606.0.0

### LINKED ISSUE #1074: Build failure on nightly Rust
Hi,

`cargo +nightly install racer` fails with the following:

```
error[E0046]: not all trait items implemented, missing: `mixed_site`
   --> /home/jakob/.cargo/registry/src/github.com-1ecc6299db9ec823/rustc-ap-syntax-583.0.0/ext/proc_macro_server.rs:657:1
    |
657 | impl server::Span for Rustc<'_> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `mixed_site` in implementation
    |
    = note: `mixed_site` from trait: `fn(&mut Self) -> <Self as proc_macro::bridge::server::Types>::Span`
```

This is on `rustc 1.40.0-nightly (2e7244807 2019-10-04)`, and the version of `rustc-ap-syntax` being brought in is `v583.0.0`.

TsarFox:
Resolves #1074

### DIFF
--- a/src/racer/lib.rs
+++ b/src/racer/lib.rs
@@ -11,6 +11,8 @@ extern crate bitflags;
 #[macro_use]
 extern crate derive_more;
 
+extern crate rustc_ap_syntax as syntax;
+
 #[macro_use]
 mod testutils;
 #[macro_use]
--- a/src/racer/lib.rs
+++ b/src/racer/lib.rs
@@ -11,8 +11,6 @@ extern crate bitflags;
 #[macro_use]
 extern crate derive_more;
 
-extern crate rustc_ap_syntax as syntax;
-
 #[macro_use]
 mod testutils;
 #[macro_use]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,13 +39,8 @@ dependencies = [
 ]
 
 [[package]]
-name = "autocfg"
-version = "0.1.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
 name = "bitflags"
-version = "1.1.0"
+version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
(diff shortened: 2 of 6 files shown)


kngwyu:
Thank you!

src/racer/lib.rs
```diff
@@ -11,6 +11,8 @@ extern crate bitflags;
 #[macro_use]
 extern crate derive_more;
 
+extern crate rustc_ap_syntax as syntax;
```
kngwyu: Do we really need this?

TsarFox: Yes. Without it, building complains about "use of undeclared type or module `syntax`"

TsarFox: I suppose the alternative is changing every reference of `syntax` to `rustc_ap_syntax`, but I thought this was a little less intrusive

kngwyu: Ah, OK.
Then, how about setting it in toml
```toml
[dependencies.syntax]
package = "rustc-ap-syntax"
version = "606.0.0"
```
like [rustfmt does](https://github.com/rust-lang/rustfmt/pull/3835/files)?

TsarFox: Didn't know you could do that, thanks :]

I agree that's much nicer. Changed.

kngwyu:
Thank you for your quick fix, again.

TsarFox:
My pleasure :)

ghedo:
@kngwyu is there an ETA for a new release with this fix? Just so https://github.com/rust-lang/rust/issues/65084 and https://github.com/rust-lang/rust/issues/65085 can be unblocked.

kngwyu:
@ghedo 
Released 2.1.28 now.
Thank you for letting me know that.

PR #1083, opened 2019-11-23
Complete async fn

### LINKED ISSUE #1055: racer-find-definition async syntax support?
I tried using racer on what I believe is a soon-to-be stable [async/await Rust code](https://github.com/rustasync/runtime/blob/master/examples/hello.rs):

Screenshot from 2019-07-09 16-58-51

when the cursor is over `say_hi().await` I cannot jump to the function definition above:
```
$ racer find-definition 9 5 say_hi src/main.rs 
END
```

If I turn that snippet into sync code, it resolves the definition:
```
$ racer find-definition 9 5 say_hi src/main.rs 
MATCH say_hi,3,3,say_hi,Function,fn say_hi()
END
```

Opinions? Is my assumption correct?

kngwyu:
Resolve #1055


apiraino:
Great job, thank you so miuch! :+1:

PR #1090, opened 2020-02-05
metadata: Don't use deprecated Error::description

Xanewok:
we will forward to `Display`, which is implemented the same way, anyway.

See https://github.com/rust-lang/rust/pull/66919


kngwyu:
Thanks!

PR #1091, opened 2020-02-11
upgrade to rustc-ap* v642

calebcartwright:
Refs https://github.com/rust-lang/rust/issues/68916
Refs https://github.com/rust-lang/rust/issues/68917

v642+ of libsyntax and friends is now needed to compile on nightly. We've updated rustfmt to use v642 (https://github.com/rust-lang/rustfmt/pull/4043) and this does the same for Racer so that Rls can subsequently be updated as well. 

This is my first contribution to Racer so please let me know if I've missed anything/need to do anything else!

cc @topecongiro @Xanewok

### DIFF
--- a/src/racer/ast_types.rs
+++ b/src/racer/ast_types.rs
@@ -15,10 +15,10 @@ use syntax::ast::{
     self, GenericBound, GenericBounds, GenericParamKind, LitKind, PatKind, TraitRef, TyKind,
     WherePredicate,
 };
+use rustc_span::source_map;
+use rustc_ast_pretty::pprust;
 // we can only re-export types without thread-local interned string
 pub use syntax::ast::{BindingMode, Mutability};
-use syntax::print::pprust;
-use syntax::source_map;
 
 /// The leaf of a `use` statement.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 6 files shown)


src/racer/ast_types.rs
```diff
@@ -159,11 +159,11 @@ impl Ty {
             LitKind::Byte(_) => make_match(PrimKind::U8),
             LitKind::Char(_) => make_match(PrimKind::Char),
             LitKind::Int(_, int_ty) => make_match(PrimKind::from_litint(int_ty)),
-            LitKind::Float(_, float_ty) => match float_ty {
+            ast::LitKind::Float(_, ast::LitFloatType::Unsuffixed) => make_match(PrimKind::F32),
```
kngwyu: Just `LitKind`?

kngwyu:
Thank you for your PR!
Almost LGTM :+1:

src/racer/ast_types.rs
calebcartwright: Oops! Copy/pasta from rustfmt sorry 😄 Do you have a preference on whether this fix should be squashed in with the original commit, or as a new commit?

kngwyu:
Thank you!

PR #1099, opened 2020-03-30
update to rustc-ap-* v651

calebcartwright:
Updates the rustc-ap-* dependencies to v651, which brings rustfmt and racer inline with the same version of those crates

https://github.com/rust-lang/rust/issues/70280
https://github.com/rust-lang/rustfmt/pull/4100

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,6 +39,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "autocfg"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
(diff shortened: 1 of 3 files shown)


calebcartwright:
Travis seems to have passed but that status isn't changing in the check :man_shrugging: 

https://travis-ci.org/github/racer-rust/racer/builds/668940086?utm_source=github_status&utm_medium=notification

kngwyu:
@calebcartwright 
> Travis seems to have passed but that status isn't changing in the check man_shrugging

I sometimes see this bug, no worry.

Thank you!

PR #1107, opened 2020-04-15
Fix #1052.

### LINKED ISSUE #1052: Completion not working with Rust 2018 `no more mod.rs`
Racer isn't able to complete modules that looks like `my_module.rs`, `my_module/sub_mod.rs` (https://doc.rust-lang.org/nightly/edition-guide/rust-2018/module-system/path-clarity.html#no-more-modrs).
Changing to `my_module/mod.rs`, `my_module/sub_mod.rs` makes it work again.

rbartlensky:
The code now looks up the required module correctly, depending on the file in
which its declaration was found.

I didn't add a test for this. Could you point me to where I could setup a crate, and test the module search?

### DIFF
--- a/src/racer/matchers.rs
+++ b/src/racer/matchers.rs
@@ -352,48 +352,74 @@ pub fn match_mod(
         });
     } else {
         debug!("found a module declaration: |{}|", blob);
-
+        // the name of the file where we found the module declaration (foo.rs)
+        // without its extension!
+        let filename = context.filepath.file_stem()?;
         let parent_path = context.filepath.parent()?;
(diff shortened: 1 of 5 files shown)


kngwyu:
Thank you!

You can add a module on [test_project](https://github.com/racer-rust/racer/tree/master/test_project/src) and test it by `with_test_project`.
See [external.rs](https://github.com/racer-rust/racer/blob/master/tests/external.rs) for how to use it.

rbartlensky:
@kngwyu Thank you for your help! I wrote some tests; I hope they're OK! Let me know if you need anything else.

kngwyu:
LGTM, thanks :100:

Issue #1119, opened 2020-07-17
Attach binary executables to releases?


fitzgen:
Sometimes building racer breaks because of nightly stuff (e.g. nightly-2020-07-16 doesn't build racer 2.1.35). It would be cool if the latest tagged releases had prebuilt binary executables for linux/macos/windows attached.

For example, `cargo-wasi` automatically builds and publishes binaries in github actions, which might be helpful to borrow from:

* https://github.com/bytecodealliance/cargo-wasi/blob/ea1da176cba5b6714b419f8c20b79ae7fb4c5d0f/.github/workflows/main.yml#L68-L118

* https://github.com/bytecodealliance/cargo-wasi/blob/ea1da176cba5b6714b419f8c20b79ae7fb4c5d0f/.github/workflows/main.yml#L150-L171

kngwyu:
Thanks!
I think it's definitely a good idea since we have many 'cannot complie' issues and I don't have enough time and energy to deal with them, though it's embarrassing.

PR #1122, opened 2020-07-21
Add rust-toolchain file

svenstaro:
This makes sure that people will always have the correct nightly build that this works with. It also makes sure the build is reproducible (as different nightly version might behave differently).


kngwyu:
Thanks, but I fear it can be outdated too quickly. Maybe binary release (#1119) is better.

svenstaro:
I still think it would make a lot of sense in terms of reproducibility for distro packagers like me. Maybe consider maintaining a `rust-toolchain` in the future. :)

Probably it doesn't get outdated _that_ quickly.

kngwyu:
OK, now I think I understand the merit and decided to merge this for now, though I'm not sure I can handle this effectively.

svenstaro:
Sadly and ironically, as it turns out, the particular version 2020-07-21 [is broken](https://github.com/rust-lang/rust/issues/74577). I'm using nightly-2020-07-20 and that works fine.

PR #1129, opened 2020-09-03
bump rustc-ap crates  to v677

calebcartwright:
We need to bump the rustc-ap crates again to fix the broken toolstates for rustfmt and rls on nightly.

AFAICT the 5 failing tests are existing/unrelated, and reproducible on master

https://github.com/rust-lang/rust/issues/76146
https://github.com/rust-lang/rust/issues/76145


Xanewok:
FWIW this doesn't seem to regress anything, assuming my #1130 actually works and doesn't green-light every test :sweat_smile: 

As such, I think it'd be great to land this asap due to broken toolstate.

@calebcartwright would you mind also updating `rust-toolchain` here to e.g. `2020-09-02`? It seems the CI is failing since we're using too old of a nightly toolchain.

Xanewok:
Thanks!

Xanewok:
Published a new version: https://crates.io/crates/racer/2.1.38

@kngwyu fyi

kngwyu:
Thanks!

PR #1130, opened 2020-09-03
Adapt to the new standard library directory layout

Xanewok:
Refs: https://github.com/rust-lang/rust/pull/73265

@kngwyu Does that look acceptable? I'm somewhat torn where should I add a new fallback and where I should just straight up replace the old path logic. I think it makes sense to wholly replace it since Racer is tied to the version of Rust that distributes the rustc-ap-* library set it depends on but I'm not sure how much we'd like to maintain backwards-compat for older Rust toolchains...


kngwyu (approved):

kngwyu:
Thanks!

> but I'm not sure how much we'd like to maintain backwards-compat for older Rust toolchains...

Since it is not *very* difficult to try both of pathes, I'll add some fallbacks for macros and premitives later.

Xanewok:
Just tested this on Windows and the first run seem to have failed but other subsequent `cargo test` runs succeed. Tried with and without `RUST_SRC_PATH` set both in Powershell and in mingw-w64. Maybe it's due to stale build cache?

src/racer/primitive.rs
```diff
@@ -4,8 +4,8 @@ use crate::nameres::{self, RUST_SRC_PATH};
 use rustc_ast::ast::{IntTy, LitIntType, UintTy};
 use std::path::PathBuf;
 
-const PRIM_DOC: &str = "libstd/primitive_docs.rs";
```
Leorii, line 7: Changing this to the new directory layout without also checking for the old layout will cause panics on any system that still has the old directory layout.

PR #1135, opened 2020-10-15
Fix the default value of RUST_SRC_PATH

xuyizhe:
Both `rust/library` and `rust/src` exist in the current version, the `std` directory is not matched correctly.

```shell
$ rustc --version
rustc 1.48.0-nightly (0da580074 2020-09-22)

$ racer --version
racer 2.1.39

$ racer complete std::io::B                         
Unable to find libstd under RUST_SRC_PATH. N.B. RUST_SRC_PATH variable needs to point to the *src* directory inside a rust checkout e.g. "/home/foouser/src/rust/src". Current value ""/Users/xuyizhe/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/src/libstd""
```


kngwyu (approved):
Thanks!

PR #1136, opened 2020-10-15
Just update docs to reflect new location of rust source.

scuellar:
Since [rust-lang/rust/pull/73265](https://github.com/rust-lang/rust/pull/73265), location of rust source files has changed and after [this merge](https://github.com/racer-rust/racer/pull/1130), the documentation in the README and other places (e.g. src/racer/util.rs) needs to reflect that the rust source is now in `...lib/rustlib/src/rust/library` instead of `lib/rustlib/src/rust/src`. 

I changed the documentation to reflect both the new location and the old one for those with old rust distributions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ As mentioned in the command output, don't forget to add the installation directo
 
 1. Fetch the Rust sourcecode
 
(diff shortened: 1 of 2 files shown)


kngwyu (approved):
Thanks!

PR #1138, opened 2020-10-25
Update README.md to use **/src/rust/library

leodutra:
`RUST_SRC_PATH` last folder was renamed from `/src` to `/library`.

refs:
https://github.com/racer-rust/racer/pull/1130
https://github.com/racer-rust/racer/issues/1137


kngwyu (approved):
Thank you!

PR #1143, opened 2020-11-28
deps: bump rustc-ap to v691

calebcartwright:
We have another broken toolstate issue with RLS and rustfmt that requires bumping the rustc-ap crates to the latest version. 

I am getting one test failure in my local environment even after these changes, but I was getting five test failures with the latest on master so I'm guessing this is unrelated

Refs https://github.com/rust-lang/rust/issues/79406 and https://github.com/rust-lang/rust/issues/79407

r? @kngwyu and/or @Xanewok


kngwyu (approved):
Thanks, I'll fix these tests later.

Xanewok:
Should we block releasing 2.1.41 on fixed tests?

kngwyu:
@Xanewok 
This failure is because the stdlib location changed and can be critical.
I'll do my best but please wait for 1~2 days (now my worktime for OSS is limited, sorry).

Xanewok:
> I am getting one test failure in my local environment even after these changes, but I was getting five test failures with the latest on master so I'm guessing this is unrelated

@calebcartwright this was probably caused by not having rust-src component - that's what I'm seeing locally after introducing the `rust-toolchain` file. One fix would be to use the new format (introduced in Rustup 1.23), i.e.:
```toml
# rust-toolchain
[toolchain]
channel = "nightly-2020-11-27"
components = ["rust-src"]
```
After installing this locally, now only `completes_for_alias_with_resolved_generics`, which is the same error that is failing in the CI.

Xanewok:
Running the test with `RUST_LOG=trace` (and calling `env_logger::try_init()`) yields a following log:
```
[2020-11-30T19:47:32Z DEBUG racer::ast] destructure_pattern_to_ty point BytePos(4) ty Match(Match ["Vec", "/home/xanewok/.rustup/toolchains/nightly-2020-11-27-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec.rs", BytePos(10934), true, Struct(GenericsArgs([])), |pub struct Vec<T, #[unstable(feature = " ", issue = " ")] A: AllocRef = Global>|]) pat: Ident(ByValue(Not), v#0, None)
```

There's a good chance that https://github.com/rust-lang/rust/pull/78461 broke the test - it seems that inline attribute on a generic parameter is not recognized, i.e. `Struct(GenericsArgs([]))`. I'll dig deeper.

Xanewok:
FYI I published a [2.1.41](https://crates.io/crates/racer/2.1.41) version that includes this dep bump.

cc @kngwyu

kngwyu:
Thanks!

PR #1144, opened 2020-12-01
Fix regressed test due to rust-lang/rust#78461

Xanewok:
See https://github.com/racer-rust/racer/pull/1143 for more context.

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -527,7 +527,16 @@ impl<'c, 's, 'ast> visit::Visitor<'ast> for ExprTypeVisitor<'c, 's> {
                                     .and_then(|ty| path_to_match(ty, self.session))
                             }
                             MatchType::Method(ref gen) => {
-                                typeinf::get_return_type_of_function(&m, &m, self.session).and_then(
+                                let mut return_ty = typeinf::get_return_type_of_function(&m, &m, self.session);
+                                // Account for already resolved generics if the return type is Self
+                                // (in which case we return bare type as found in the `impl` header)
+                                if let (Some(Ty::Match(ref mut m)), Some(gen)) = (&mut return_ty, gen) {
+                                    let resolved = gen.args().filter_map(|tp| tp.resolved());
+                                    for (l, r) in m.generics_mut().zip(resolved) {
+                                        l.resolve(r.clone());
+                                    }
+                                }
+                                return_ty.and_then(
                                     |ty| {
                                         path_to_match_including_generics(
                                             ty,
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -531,9 +531,10 @@ impl<'c, 's, 'ast> visit::Visitor<'ast> for ExprTypeVisitor<'c, 's> {
                                 // Account for already resolved generics if the return type is Self
                                 // (in which case we return bare type as found in the `impl` header)
                                 if let (Some(Ty::Match(ref mut m)), Some(gen)) = (&mut return_ty, gen) {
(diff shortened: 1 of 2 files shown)


src/racer/ast.rs
```diff
@@ -527,7 +527,16 @@ impl<'c, 's, 'ast> visit::Visitor<'ast> for ExprTypeVisitor<'c, 's> {
...
                             }
                             MatchType::Method(ref gen) => {
-                                typeinf::get_return_type_of_function(&m, &m, self.session).and_then(
+                                let mut return_ty = typeinf::get_return_type_of_function(&m, &m, self.session);
+                                // Account for already resolved generics if the return type is Self
+                                // (in which case we return bare type as found in the `impl` header)
+                                if let (Some(Ty::Match(ref mut m)), Some(gen)) = (&mut return_ty, gen) {
+                                    let resolved = gen.args().filter_map(|tp| tp.resolved());
+                                    for (l, r) in m.generics_mut().zip(resolved) {
+                                        l.resolve(r.clone());
```
Xanewok, line 536: This looks like a hack but I encountered a similar resolve/substitution code in other places, so it seems it's not as out of place as it looks at a first glance.

There may be other designs how to tackle this but decided for it because it's least invasive for now.

kngwyu, line 536: Can this cause the order mismatch?
```suggestion
                                    for (type_param, arg) in m.generics_mut().zip(gen.args()) {
                                        if let Some(resolved) = arg {
                                            type_param.resolve(resolved.clone());
```

kngwyu (approved):
Thanks!

src/racer/ast.rs
Xanewok: In terms of order itself, I saw that the generics/type parameters are collected left-to-right in the AST order, so I don't think it's possible, at a first glance, to mix up the order itself.

I'm not so well versed in Racer internals, but what I found somewhat surprising at first is that `MatchType::Method(ref gen)` contains generics for the calling type rather than method itself. It looks odd at first' I'd expect it to contain only `G` in the following example, e.g.
```rust
impl<T> MyGenType<T> {
  pub fn another_generic<G>() { ... }
}
```
rather than (resolved) `T`.
I **didn't test** that but I'd expect the above method `Match` to contain generics as `[T, G]` or `[T]` - if we have some type information about the `T` type or also `G`, it shouldn't hurt to zip and resolve any additional type information we have at the call site. I don't think the `Match` would contain `[G, T]` - that'd be *very* surprising (and the order would be mixed up, yes).

In the regressed test the method
```rust
impl<T> Vec<T> {
  pub const fn new() -> Self { ... }
```
There are no type parameters in the method itself, but the `MatchType::Method(ref gen)` contains generics for `[T, A]` that `Vec` is generic over, with the `T = String` being resolved at call site. Since `get_return_type_of_function` returns with unresolved generics `[T, A]`, we just zip it with the generics we have.

I hope that's somewhat clear, let me know if I should rephrase that :sweat_smile:

kngwyu: I'm worrying that `filter_map` can cause the order mismatch when the `resolved` returns None

Xanewok: Oh, now I see what you mean. I operated under the impression that there are `[resolved, ..unresolved]` type parameters but that's probably incorrect if we had `Vec<_, Global>` (I'm not sure when or if at all type inference kicks). I'll commit your suggestion, thanks!

Xanewok:
Squashed the last few fixup commits and force-pushed

Xanewok:
Going to merge the final version since it's approved. Thanks for taking a look!

kngwyu:
Thanks!

PR #1146, opened 2020-12-02
Add GitHub Actions CI

### LINKED ISSUE #1145: Consider switching CI to GitHub Actions
Right now we have
- AppVeyor builds for testing `{x86_64, i686}-pc-windows-msvc` and
- Travis CI for `x86_64-unknown-linux-gnu`

which requires maintaining two config files and integrations; I also vaguely recall some problems when using AppVeyor, e.g. now it fails on missing rust-src in https://ci.appveyor.com/project/TedDriggs/racer-xr5g5/builds/36589888/job/fj30kkmy8hx7053c but Travis CI works without problems.

@kngwyu Would you accept a PR that sets up GitHub Actions for both of these platforms? We could also schedule nightly builds to catch breakage early, e.g. see https://github.com/rust-lang/rust-semverver/commit/80fac90ceabccf8a10f2eeab4cc59cd9a9de9a43 what's done in rust-semverver.

Xanewok:
Builds on #1144 (now merged) so that the tests pass.

Fixes #1145

The spin-up time is very fast and a single build runs 3-5 minutes (all 4 builds run in parallel), and so that's an improvement over what we have: AppVeyor runs 5-7 mins, Travis runs 15-20 mins (but needs some time to spin up). This also adds testing for macOS.

See https://github.com/Xanewok/racer/actions/runs/396511445 for a successful run.

r? @kngwyu


Xanewok:
I believe the action will be only triggered and run once it's committed to stop potentially malicious actors from extracting CI secrets and so on via PRs.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  push:
+    branches:
+      - master
+      - '*'
+  schedule:
+    - cron: '0 0 * * *' # Nightly at 00:00 UTC
```
kngwyu, line 11: Nice :+1: (as our build is often broken :sweat_smile:

kngwyu (approved):
Thanks!
Could please also remove `.appveyor.yml`?
Since now we run tests for Windows on actions, it is no longer useful.

Xanewok:
> Could please also remove `.appveyor.yml`?

Done.
Do you want me to remove `.travis.yml` as well, if we're removing old CI setup here?

kngwyu:
> Do you want me to remove .travis.yml as well, if we're removing old CI setup here?

Yeah, but let's remove it after watching the actions work. Thanks!

PR #1147, opened 2020-12-04
Remove .travis.yml

Xanewok:
Follow up to https://github.com/racer-rust/racer/pull/1146#issuecomment-738834770.

We don't need to merge this immediately but I'm opening it now not to forget about it.


Xanewok:
@kngwyu I fixed some remaining GHA issues in #1148 (it also took ~40 mins to just spin up a Travis run there :sob: ) so I'll merge this if you don't mind; we can always easily revert if we want to bring Travis back

PR #1148, opened 2020-12-04
Build and test all workspace packages in CI

Xanewok:
Follow-up to #1146

I lost the `--all` option when migrating, sorry about that. Will merge once CI is green.


kngwyu:
Thank you!

PR #1151, opened 2021-01-02
bump rustc-ap crates to v697

calebcartwright:
RLS and rustfmt need some recent changes upstream in rustc which require a bump of these crates. Updating them here to keep things in sync

r? @kngwyu and/or @Xanewok


kngwyu (approved):
Thanks!
I'll release a new version soon.

kngwyu:
Published 2.1.42 now.

PR #1152, opened 2021-01-02
Add support for Rust 2021.

m-ou-se:
This adds support for Rust 2021.

For now, Rust 2021 is identical to Rust 2018.


kngwyu (approved):
Thanks!

PR #1153, opened 2021-02-03
bump rustc-ap crates to v702.0.0

calebcartwright:
We've another broken toolstate issue on nightly that requires us to bump the crates in rustfmt, this does the same for racer

https://github.com/rust-lang/rust/issues/81583
https://github.com/rust-lang/rust/issues/81582


kngwyu (approved):
Thanks!

kngwyu:
Released 2.1.43 now.

PR #1161, opened 2021-02-11
Bump rustc-ap crates to version 706

thomaseizinger:
This is a follow-up from bumping the crates to version 706 in rustfmt: https://github.com/rust-lang/rustfmt/pull/4694

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -604,18 +604,18 @@ dependencies = [
 
 [[package]]
 name = "rustc-ap-rustc_arena"
-version = "705.0.0"
+version = "706.0.0"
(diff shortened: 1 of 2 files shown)


kngwyu (approved):
Thanks!
Do we need the release corresponding to this change?

thomaseizinger:
> Thanks!
> Do we need the release corresponding to this change?

Not from my side no and I think @calebcartwright wants to wait with another release dance until next week :)

calebcartwright:
@kngwyu - yes we'll need a release, but it's not terribly urgent (no broken toolstate this time) so whenever you can get around to it would be great!

Issue #1173, opened 2021-12-08
Failed to build using latest nightly


t-e-r-m:
```bash
$ cargo +nightly install racer 
    Updating crates.io index
  Installing racer v2.1.48
   Compiling libc v0.2.109
   ...
   Compiling rustc-ap-rustc_data_structures v722.0.0
   Compiling rustc-ap-rustc_arena v722.0.0
   Compiling rustc-ap-rustc_span v722.0.0
error[E0635]: unknown feature `thread_local_const_init`
  --> /home/sys2/.cargo/registry/src/github.com-1ecc6299db9ec823/rustc-ap-rustc_span-722.0.0/src/lib.rs:23:12
   |
23 | #![feature(thread_local_const_init)]
   |            ^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0635`.
error: could not compile `rustc-ap-rustc_span` due to previous error
warning: build failed, waiting for other jobs to finish...
error: failed to compile `racer v2.1.48`, intermediate artifacts can be found at `/tmp/cargo-installf8feJG`

Caused by:
  build failed
```

```bash
$ rustup run nightly rustc --version
rustc 1.59.0-nightly (0b6f079e4 2021-12-07)
```

kunfan96:
same as you,who can help me?

fakegermano:
It's as @t-e-r-m said, an upstream issue with rustc-ap-rustc_span's build pipeline, it's been stale for more than 60 days and github decided to not run any more release actions! :+1: 

If you need racer working, you can install it with `nightly-2021-06-07` (maybe a further version as well, but this is the one I tested based on approximate dates for releases.

primeNumberAndMe:
> It's as @t-e-r-m said, an upstream issue with rustc-ap-rustc_span's build pipeline, it's been stale for more than 60 days and github decided to not run any more release actions! +1
> 
> If you need racer working, you can install it with `nightly-2021-06-07` (maybe a further version as well, but this is the one I tested based on approximate dates for releases.
  
```
Installing /home/steve/.cargo/bin/racer
   Installed package `racer v2.1.48` (executable `racer`)
➜  rust-in-action racer --version
racer 2.1.48
```

 `nightly-2021-01-30-x86_64-unknown-linux-gnu` does not work with this, but `nightly-2021-06-07` is totally ok.
Thanks buddy

calebcartwright:
[...]

PR #1174, opened 2021-12-13
Use rustc_private and bump nightly

flip1995:
This moves away from the rustc-ap crates and uses rustc-dev instead. It also bumps this crate to use the latest nightly.

I just experimented with this. Since the diff was so small, I decided to submit it as a PR. Feel free to close this, if you don't want to do this.

This would help with rust-lang/rust#91543

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,33 +4,21 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.10"
+version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8716408b8bc624ed7f65d223ddb9ac2d044c0547b6fa4b0d554f3a9540496ada"
+checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
 dependencies = [
  "memchr",
 ]
 
 [[package]]
-name = "annotate-snippets"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d78ea013094e5ea606b1c05fe35f1dd7ea1eb1ea259908d040b25bd5ec677ee5"
-
-[[package]]
 name = "ansi_term"
-version = "0.11.0"
+version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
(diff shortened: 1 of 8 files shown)


calebcartwright:
If we were to go this route (and fwiw I'd like to see this change too) I think it might be helpful to update the top of the readme file and/or installation section to detail the extra requirements.

There's historically been duplicative issues (e.g. #1173) with folks attempting to install from crates.io using an unsupported toolchain, so I think emphasizing that in documentation along with the newly required `rustc-dev` component would be helpful.

One thing we heard after making the same switch in rustfmt is that the rustc-dev component isn't available for every target which was problematic for at least one case where someone was attempting to consume rustfmt as a lib. I doubt that'll be a problem here though given the nature of racer

flip1995:
I can adapt the documentation, if this should move forward.

> One thing we heard after making the same switch in rustfmt is that the rustc-dev component isn't available for every target

Oh, we didn't hear anything like that in Clippy :thinking: Going through the [components history](https://rust-lang.github.io/rustup-components-history/x86_64-unknown-linux-gnu.html), it is available on ever Tier 1 target and by random clicking on Tier 2 targets it is also available on every Tier 2 target, where also tools like rustfmt, clippy, ... are available. So I'm not that concerned here.

calebcartwright:
I think someone was trying to use rustfmt in a wasm context perhaps? Can't recall the exact specifics but there was some edge case that was surfaced to us

flip1995:
Ok, I've updated the README with a note that `rustc-dev` has to be installed to compile Racer.

I'm not that concerned, that something like this will happen to Racer. Especially since it has a note that it is not actively maintained.

calebcartwright:
@kngwyu what are your thoughts on this? The impacts of the current AP versions are starting to snowball, so think we'll either need to go this route or at least get the AP crate versions bumped

kngwyu (approved):
Thank you!

@calebcartwright 
This way is better since we don't have much maintenance ability now...

flip1995:
Should I bump this PR to the latest nightly before merging? Is anything missing, before this can get merged?

Xanewok:
@flip1995 if that would not be a problem, please update to the newest nightly; I'll merge this immediately after. Thanks for all the work here!

flip1995:
@Xanewok Just required a rust-toolchain.toml bump and a `Cargo.lock` update. 

I can also do the RLS update once a new Racer version is released :+1:

flip1995:
@Xanewok not sure if you're already at it, but are you planning to bump the Racer version soon, so that we can update Racer in RLS?

PR #1177, opened 2022-02-22
Use Lrc over Rc

Mark-Simulacrum:
This is needed for RLS to build on beta upstream (https://github.com/rust-lang/rust/pull/94228), though will likely be worked around for now to avoid delaying beta promotion.

### DIFF
--- a/src/racer/ast.rs
+++ b/src/racer/ast.rs
@@ -7,7 +7,6 @@ use crate::nameres;
 use crate::typeinf;
 
 use std::path::Path;
-use std::rc::Rc;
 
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
@Xanewok or @kngwyu -- not sure who else has permissions here -- is there something I can do to help unblock merging this? It might be a good idea to try and add some release team members or otherwise expand the folks who can merge/release here, if possible, so that we don't need to ask you to bump :)

Xanewok:
Added you with write permissions. Would it be possible to move this repository under rust-lang to avoid further breakages/reduce the release bottleneck?

PR #1180, opened 2022-04-05
Fx breakage caused by the introduction of diagnostic translations

pietroalbini:
This fixes the breakage caused by https://github.com/rust-lang/rust/pull/95512, which is going to be in *tomorrow*'s nightly. I expect CI to fail today due to that, but opening the PR in the meantime.


Xanewok:
Thanks!

PR #1183, opened 2022-05-16
Fix breakage from upstream changes in Rust

Mark-Simulacrum:
cc @Xanewok -- if we could also publish a new release, that would be great


emilyalbini:
You also need to update the pinned nightly in https://github.com/racer-rust/racer/blob/master/rust-toolchain.toml for CI to go green.

Mark-Simulacrum:
https://github.com/rust-lang/rls/pull/1775 is likely going to supercede this PR by inlining racer into RLS, so presuming that moves ahead will likely skip pushing this forward and close it.

